PR #11, opened 2018-01-11
Initial support for JIRA wiki rendering

cctile:
This PR includes support for rendering JIRA wiki markup as specified in <https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=all>

An additional script `md2jira.py` will convert a Markdown (CommonMark) formatted file to JIRA wiki and can be installed in the `bin` directory of the user's home.

### DIFF
--- a/contrib/Makefile
+++ b/contrib/Makefile
@@ -1,0 +1,8 @@
+
+INSTALL_DIR=$(HOME)/bin
+
+install: install-md2jira
+
+install-md2jira:
+	cp md2jira.py $(INSTALL_DIR)/md2jira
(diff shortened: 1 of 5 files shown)


miyuchina:
Merged in master, thanks @cctile !

I might be merging the plugins and contrib directories some time soon, and since the JIRA feature has two files, it'll have its own directory.

Also, if you want to hack on this more, could you make changes based on the dev branch, so that I can do a fast-forward merge next time? The dev branch is where most developments happen for me, so you can also see if I've made conflicting changes.

Again, thanks for the contribution!

PR #18, opened 2018-01-15
Fix for tiny import bug

huettenhain:
This fixes the import because currently, `jira_renderer.py` is in `/contrib` and there is no `/plugins` directory.


miyuchina:
Derp, sorry. Merged!

PR #22, opened 2018-01-29
RawText token for unescapables

liuq:
Hi all,

I am playing around with a literate programming tool mixing markdown and different programming languages and I discovered that, currently, the code blocks (including fences) and code inlines underwent the render_raw_text() method which could escape characters inside.
Obviously this is a wrong behaviour in case of C-like languages where ampersands `&` and curly brackets `{}` are ubiquitous. 

Therefore, for handling such a case I slightly modified the `RawText` token class including a further information `self.escape` that states whether the raw text could (when `True`) or could not (when `False`) be escaped. Also the LaTeX renderer has been modified accordingly (I forgot to modify the HTML renderer but I will do it after finishing the pull request).


liuq:
Also the HTML renderer now checks whether the raw text has to be escaped or not.

miyuchina:
Hi, sorry about the tests—those are mostly regression tests but they could be hostile to any spec changes!

I’ll review your changes when I get back. If fixing tests is too tedious, don’t feel obligated—I’ll merge it and see what I can do!

liuq:
I tried to fix some of them, however for the regression in block_token I didn't succeed (as for the mock).

miyuchina:
Thanks for the pull request! I agree with your reasoning about escaping curly brackets and ampersands. Those are definitely oversights on my part, and I'll work on patching it for the next release.

However, I have a couple of thoughts about implementing this patch:

<ol>
<li><p>It seems to me that the problem concerns rendering exclusively: we only think about whether we should escape characters when we are in the rendering phase, not when we are constructing the AST. For this reason, I think the <code>escape</code> attribute should not be introduced at token level, but instead at renderer level.</p></li>
<li><p>I also think this is a problem that only affects LaTeX. The mistake I made was that I was trying to escape characters in the <code>lstlisting</code> environment, where I didn't need to. So we can pass in the information only when we need it, for example:</p>

```python
class LaTexRenderer(BaseRenderer):
    def render_block_code(self, token):
        inner = self.render_raw_text(token.children[0], escape=False)
        ...

        return self.escape(token.content) if escape else token.content

    @staticmethod
    def escape(text):
        pass
```

<p>For HTML, we should escape characters regardless of whether they are in code blocks. Here's an improper example:</p>

```html
<pre><code>
    PrimeCalc calc = new PrimeCalc();
    Func<int, int> del = calc.GetNextPrime;
</code></pre>
```

<p>Even though I think most browsers render <code>&lt;&gt;</code> symbols correctly, they should still be escaped so as not to be interpreted as HTML tags:</p>

```html
<pre><code>
    PrimeCalc calc = new PrimeCalc();
    Func&lt;int, int&gt; del = calc.GetNextPrime;
</code></pre>
```

<p>This is why I think the issue you raised is a LaTeX-specific problem.</p>
</li>
</ol>

I suspect limiting the changes to only `LaTeXRenderer` would also be easier to implement. Do you think you'd be interested in implementing it? If not, I can do it, and we can close this pull request and move the discussion to an issue instead.

Let me know what you think!

mistletoe/html_renderer.py
```diff
@@ -78,7 +78,10 @@ def render_escape_sequence(self, token):
 
     @staticmethod
     def render_raw_text(token):
-        return html.escape(token.content)
+        if getattr(token, 'escape', True):
```
miyuchina: Is there a reason for using a `getattr` here, instead of `token.escape`? It seems that in your implementation, all `RawText` tokens will have the `escape` attribute. Or is it a test-related change?

liuq:
Yes, you're right... it's a matter of transformation not of representation. I will just fix the LaTeXrenderer then.

I ask you to drop this PR and I will issue a new one with just that modification.

Thanks and all the best,

Luca

PR #23, opened 2018-01-30
Escape LaTeX renderer *almost* fixed

liuq:
As of your comment, I'd just corrected the LaTeX render to handle correctly the code (not) escaping.
In addition, as in my previous issue I came up with a possible logo featuring the real mistletoe, look whether it is ok with you.

All the best,

Luca


liuq:
Unfortunately tests still do not pass, because using `self.render_raw_text(token.children[0], False)` breaks the mock-up in the unit testing that relies on `self.render_inner()` which is not called in this case.
What shall I do?

All the best,

Luca

miyuchina:
Thank you! I cherry-picked your commit d636a18 for `LaTeXRenderer` onto the dev branch. Tests are fixed in ec72865.

I want to keep this pull request clean and review the icon changes in another thread, so I'm closing this one for now.

PR #25, opened 2018-01-31
Headline fix

huettenhain:
The newline at this point is unnecessary and for example, it creates a line break at the beginning of each document that begins with a headline.


miyuchina:
I'm not well-versed enough with JIRA syntax, so I was wondering if @cctile could look through this.

miyuchina:
Hmm. Since @cctile is busy, I'm going to merge this for now.

PR #27, opened 2018-02-13
Read from stdin in md2jira if no input file is given

alexkolson:
If no input file is given, read from stdin.


miyuchina:
Sweet, thank you so much!

PR #57, opened 2018-09-02
Fixed #54: loose list when followed by other token

### LINKED ISSUE #54: lists should not contain <p>
Why do you use &lt;p&gt;&lt;/p&gt; in lists ? This is not the way Commonmark does it.

Other strange behavior: in the same document, some lists get the paragraphs, some others don't. Could you explain how it is supposed to work ?

Rogdham:
This patch seems to fix #54, but please review carefully this pull request to be sure that there are no side-effects: I don't really know your codebase.

New test case added as well to test against future regression on the issue.


miyuchina:
This is awesome, thank you!

PR #59, opened 2018-09-06
Fix FileWrapper backstep after StopIteration

### LINKED ISSUE #58: FileWrapper backstep after StopIteration
When you reach the end of a `FileWrapper`, continuing to call `__next__` still increases `_index`. As a result, you would need to call `backstep` that many times.

Example code:
```python
lines = FileWrapper(['a', 'b'])

for line in lines:
    print(line)  # will print 'a' then 'b'

lines.backstep()

# at that point next(lines) still raise StopIteration

# so you need an extra call to backstep if you want to get the last line
lines.backstep()  #  <--- this line is what the issue is about

next(lines)  # gives 'b'
```

For me that's one call too many to `backstep`.

Moreover, if the `StopIteration` is raised more than once, you need to call `backstep` that many times (which means that you need to know how many times the exception was raised).

**Is the current behaviour the intended one, or is it open to be changed?**

***

Below is a suggested patch to try the suggested behaviour:
```patch
--- a/mistletoe/block_tokenizer.py
+++ b/mistletoe/block_tokenizer.py
@@ -10,8 +10,8 @@ class FileWrapper:
         self._anchor = 0
 
     def __next__(self):
-        self._index += 1
-        if self._index < len(self.lines):
+        if self._index + 1 < len(self.lines):
+            self._index += 1
             return self.lines[self._index]
         raise StopIteration
```

_Note: all tests are still passing with the patch applied._

Rogdham:
This PR fix #58: see that issue for the rationale.

### DIFF
--- a/mistletoe/block_tokenizer.py
+++ b/mistletoe/block_tokenizer.py
@@ -10,8 +10,8 @@ class FileWrapper:
         self._anchor = 0
 
     def __next__(self):
-        self._index += 1
-        if self._index < len(self.lines):
+        if self._index + 1 < len(self.lines):
+            self._index += 1
             return self.lines[self._index]
         raise StopIteration
 


miyuchina:
Thank you!

PR #62, opened 2018-09-07
Allow more than one level of token subclass

Rogdham:
Currently you subclass a token and then subclass that token…

For example:

```python
class Foo(SpanToken):

    pass


class Bar(Foo):

    pattern = re.compile('(@@@)')


class BarRenderer(HTMLRenderer):

    def __init__(self):
        super().__init__(Bar)


mistletoe.markdown('', BarRenderer)
```

gives the following error:
```
  File "bug.py", line 21, in <module>
    mistletoe.markdown('', BarRenderer)
  File "/path/to/mistletoe/lib/python3.7/site-packages/mistletoe/__init__.py", line 18, in markdown
    with renderer() as renderer:
  File "bug.py", line 18, in __init__
    super().__init__(Bar)
  File "/path/to/mistletoe/lib/python3.7/site-packages/mistletoe/html_renderer.py", line 30, in __init__
    super().__init__(*chain((HTMLBlock, HTMLSpan), extras))
  File "/path/to/mistletoe/lib/python3.7/site-packages/mistletoe/base_renderer.py", line 76, in __init__
    inspect.getmodule(token.__bases__[0]).add_token(token)
AttributeError: module '__main__' has no attribute 'add_token'

```

This PR offers to detect if a token is span or block based on subclass of the `SpanToken` instance instead of using the `inspect` module.

### DIFF
--- a/mistletoe/base_renderer.py
+++ b/mistletoe/base_renderer.py
@@ -4,7 +4,6 @@ Base class for renderers.
 
 import re
 import sys
-import inspect
 from mistletoe import block_token, span_token
 
 class BaseRenderer(object):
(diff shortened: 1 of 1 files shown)


miyuchina:
Thank you!

Issue #65, opened 2018-09-20
Unexpected Paragraph Tag in Lists


href:
When using an ordered and an unordered list in a single markdown, unexpected paragraphs are used inside the list items:

```python
from mistletoe import markdown
markdown('* foo\n\n1. bar')
```
Result:
```html
<ul>
    <li>
        <p>foo</p>
    <li>
</ul>
<ol>
    <li>bar</li>
</ol>
```

If the lists are rendered separately, no paragraph is used. Almost looks like there exists some shared state.

miyuchina:
Should be fixed by 5191b42.

PR #66, opened 2018-10-22
Fixed incorrect loose list (#65)

Vallentin:
This pull request fixes the issue mentioned in #65.

Markdown:

```markdown
* Foo

1. Bar
```

Before:

```html
<ul>
    <li>
        <p>Foo</p>
    <li>
</ul>
<ol>
    <li>Bar</li>
</ol>
```

After:

```html
<ul>
    <li>Foo<li>
</ul>
<ol>
    <li>Bar</li>
</ol>
```

### DIFF
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -477,6 +477,12 @@ class List(BlockToken):
             matches.append(output)
             if next_marker is None:
                 break
+
+        if matches:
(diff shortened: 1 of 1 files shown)


MrVallentin:
Moved the check into `List.read`, as it seems more fitting to be part of the parsing stage instead.

PR #69, opened 2018-11-08
Include `span` and `svg` in the span token tags

tonyfast:
I thought these would be useful tags to add the span token tags set for discovering inline html.


miyuchina:
I'm currently not including `span` or `svg` when parsing `HTMLSpan` because the CommonMark spec does not include them (see the spec for [HTML block](https://spec.commonmark.org/0.29/#html-blocks)). Let me know if there are further questions, but I think it should be relatively easy to patch `HTMLSpan` in your own code to include support for these tags?

PR #72, opened 2018-11-15
Added parent attribute

### LINKED ISSUE #71: Get parent token
I've implemented a custom `HTMLRenderer`, where I want to know if `render_image` is the only element on a paragraph.

Essentially be able to differentiate the output between these:

```markdown
Foo ![](foobar.png) Bar

![](foobar.png)
```

This could be possible if `token` has a `parent` attribute.

```python
class TestRenderer(HTMLRenderer):
    def render_image(self, token):
        has_more = len(token.parent.children) > 1
        if has_more:
            ...
        else:
            ...
```

Currently I have a `_set_parent` function, that is applied to `Document` in `render`.
However it would be nice, if this was part of mistletoe and done as children are added to the token.

```python
from mistletoe import HTMLRenderer
from mistletoe.block_token import Document

def _set_parent(parent):
    if hasattr(parent, "children"):
        for child in parent.children:
            setattr(child, "parent", parent)
            _set_parent(child)

class TestRenderer(HTMLRenderer):
    def render(self, token):
        if isinstance(token, Document):
            _set_parent(token)
        return super().render(token)

    def render_image(self, token):
        has_more = len(token.parent.children) > 1
        ...
```

vallentin:
Resolves #71.

### DIFF
--- a/mistletoe/ast_renderer.py
+++ b/mistletoe/ast_renderer.py
@@ -38,4 +38,6 @@ def get_ast(token):
         node['header'] = get_ast(node['header'])
     if 'children' in node:
         node['children'] = [get_ast(child) for child in node['children']]
+    if 'parent' in node:
+        node['parent'] = node['parent'].__class__.__name__
     return node
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -121,6 +121,8 @@ class BlockToken(object):
     """
     def __init__(self, lines, tokenize_func):
         self.children = tokenize_func(lines)
+        for child in self.children:
+            child.parent = self
 
     def __contains__(self, text):
         return any(text in child for child in self.children)
@@ -148,6 +150,8 @@ class Document(BlockToken):
         _root_node = self
         span_token._root_node = self
         self.children = tokenize(lines)
+        for child in self.children:
+            child.parent = self
         span_token._root_node = None
         _root_node = None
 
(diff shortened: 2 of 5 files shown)


Vallentin:
*Here's a suggestion.*

Instead of all the `for child in children`, it might be beneficial to change `__init__` of `SpanToken` and `BlockToken`. To allow and instead pass all children via `super()`.

Something like:

```python
class SpanToken:
    ...

    def __init__(self, match, children=None):
        if not self.parse_inner:
            self.content = match.group(self.parse_group)
        if children is not None:
            self.children = children
            for child in self.children:
                child.parent = self

    def __contains__(self, text):
        if hasattr(self, 'children'):
            return any(text in child for child in self.children)
        return text in self.content
```

What do you think @miyuchina? Is the `parent` attribute even something you want in the first place?

pbodnar:
This looks promising. I totally agree with some refactoring though, so that the logic of setting `parent` into `children` is kept just at one place.

pbodnar:
Hi there, I would like to finally finish this feature after all the years. :) I think now, the simplest thing to do is to make the `children` attribute a `@property` with a setter function which would take care of setting the `parent` to all the `children`. Now that we've already got the common `Token` ancestor class, it would mean extending just this class. I will check this out and possibly create a fresh PR...

pbodnar:
Replaced by #206. Thanks for the initial idea. :)

Issue #74, opened 2018-12-21
HTML output without inline HTML blocks


skairunner:
Markdown that doesn't accept inline HTML is a common use-case for comments and as a quick way to sanitize input. A contrib renderer that renders HTML while escaping inline HTML tags would be convenient to have.

I'm trying to implement this myself--unless my initial tests are wrong, it seems I can simply inherit from HTMLRenderer and override the `render_html_block` and `render_html_span` methods--and I'll have a PR up soonish.

skairunner:
One downside of the current method is that any line or lines that count as an HTMLDiv do not have a wrapping <p>.

```html
: input
<div>hello</div>

: expected 
<p>&lt;div&gt;hello&lt;/div&gt;</p>

: current
&lt;div&gt;hello&lt;/div&gt;
```

This doesn't happen to tokens counted as span. I'm not sure what the best way to implement wrapping <p> on HTMLBlocks is. (1) is the quick and simple appending/prepending <p> manually on the `LimitedHTMLRenderer.render_html_block` method, but I wonder if this could cause problems down the line. (2) is using different tokens that counts all html as Span level tokens instead of separating some as Block. Thoughts?

Finally, the current tests feel a little brittle. Any suggestions on improvements would be really appreciated.

pbodnar:
Or, as an alternative and hopefully working solution, what about simply disable parsing of HTML elements? I. e. change the following line in your `HTMLRenderer` version:

    super().__init__(*chain((HTMLBlock, HTMLSpan), extras))

... to:

    super().__init__(*chain((), extras))

The existing `HTMLRenderer` could possibly take a boolean init parameter that would encapsulate this switching.

What do you think?

PR #75, opened 2018-12-21
💡 Contrib limited html renderer (#74)

skairunner:
As per #74


pbodnar:
@skairunner, thank you for your contribution. As suggested in #74, I've prepared an alternative solution at #204.

PR #93, opened 2020-03-01
Latex renderer escaping underlines

sebinq:
During LaTeX conversion we do need to escape underlines from output (besides it is InlineBlock token's content)

### DIFF
--- a/mistletoe/latex_renderer.py
+++ b/mistletoe/latex_renderer.py
@@ -23,7 +23,12 @@ class LaTeXRenderer(BaseRenderer):
         return '\\textit{{{}}}'.format(self.render_inner(token))
 
     def render_inline_code(self, token):
-        return '\\verb|{}|'.format(self.render_inner(token))
+        content = ""
+        # there should be only one children...
+        for children in token.children:
+            content += self.render_raw_text(children, escape=False)
+
+        return '\\verb|{}|'.format(content)
 
     def render_strikethrough(self, token):
         self.packages['ulem'] = ['normalem']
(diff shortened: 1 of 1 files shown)


mistletoe/latex_renderer.py
```diff
@@ -23,7 +23,12 @@ def render_emphasis(self, token):
         return '\\textit{{{}}}'.format(self.render_inner(token))
 
     def render_inline_code(self, token):
-        return '\\verb|{}|'.format(self.render_inner(token))
+        content = ""
+        # there should be only one children...
+        for children in token.children:
+            content += self.render_raw_text(children, escape=False)
+
+        return '\\verb|{}|'.format(content)
```
pbodnar, line 31: OK, but when talking about (not) escaping, we should also cope with `|` here? Please see https://tex.stackexchange.com/questions/10629/escape-in-verb: It looks like one can choose nearly any character as the delimiter, but the character must not be present within the rendered text.

mistletoe/latex_renderer.py
```diff
@@ -51,9 +56,9 @@ def render_escape_sequence(self, token):
         return self.render_inner(token)
 
     def render_raw_text(self, token, escape=True):
-        return (token.content.replace('$', '\$').replace('#', '\#')
-                             .replace('{', '\{').replace('}', '\}')
-                             .replace('&', '\&')) if escape else token.content
+        return (token.content.replace('$', '\\$').replace('#', '\\#')
```
pbodnar, line 59: These double escaped back-slashes are not actually necessary here for Python, are they? I propose not to change these lines then.

mistletoe/latex_renderer.py
```diff
@@ -51,9 +56,9 @@ def render_escape_sequence(self, token):
...
+                             .replace('{', '\\{').replace('}', '\\}')
+                             .replace('&', '\\&')).replace('_', '\\_') if escape else token.content
```
pbodnar, line 61: Yes, this seems OK, but what `%`, it looks like it should be escaped as well. I'm no expert on LaTeX though, so don't hesitate to correct me.

pbodnar (changes requested):
@sebqq, thanks for your PR. It looks generally OK, but I've got some questions which I would be glad if you could answer. It would be also great if you could extend corresponding unit tests in order to cover your changes.

PR #97, opened 2020-04-22
JIRARenderer: fix list items to have a single newline in between each instead of two

dgoldstein0:
this change is important especially for numbered lists, because with 2 lines between items in numbered lists, jira interprets that as each item being a separate list, and restarts numbering from 1.

Tested
```
    $ python3
    >>> from mistletoe.contrib.jira_renderer import JIRARenderer
    >>> import mistletoe
    >>> mistletoe.markdown('1. step one\n2. step two\n3. step three\n', JIRARenderer)
    '# step one\n# step two\n# step three\n'
    >>> mistletoe.markdown('1. step one\n2. step two\n    - blah\n    - blah\n3. step three\n', JIRARenderer)
    '# step one\n# step two\n#* blah\n#* blah\n# step three\n'
    >>> mistletoe.markdown('1. step one\n2. step two\n    1. blah\n    2. blah\n3. step three\n', JIRARenderer)
    '# step one\n# step two\n## blah\n## blah\n# step three\n'
```

### DIFF
--- a/contrib/jira_renderer.py
+++ b/contrib/jira_renderer.py
@@ -118,7 +118,7 @@ class JIRARenderer(BaseRenderer):
         return inner
 
     def render_list_item(self, token):
-        template = '{prefix} {inner}\n'
+        template = '{prefix} {inner}'
         prefix = ''.join(self.listTokens)
         result = template.format(prefix=prefix, inner=self.render_inner(token))
         return result


dgoldstein0:
btw I'd be happy to contribute to the test suite - already wrote my manual test plan into some actual tests, just didn't know where to add them to code in here.

pbodnar:
@dgoldstein0, you are correct. But there are also other major problems in correctly handling new lines for correct Jira output which I'm about to share in my fork.

Unfortunately, it looks like the original author(s) has no longer time for this project, so I'm not sure if it pays off to do a PR here - @miyuchina, is it worth?

pbodnar:
Hi again, I have the write access now. :) So I merged #100 which includes the change from this PR as well.

PR #105, opened 2020-09-11
Fix jira_renderer to correctly output table headers; add a test.

dmitrym0:
A small fix; without it the first (header) row was replaced with the second row.


pbodnar (approved):
Hi, this looks good, thanks for the fix and the test. 👍

pbodnar:
I have merged this PR within #100. Thank you, @dmitrym0.

PR #112, opened 2021-09-14
 LaTeXRenderer: fix escaping on % sign

jasondenton:


jasondenton:
approved

pbodnar:
@jasondenton, good catch, thank you. It looks like the original author(s) simply overlooked escaping of this `%` character, didn't they? But what about `_`, shouldn't it be escaped as well? I base this on https://tex.stackexchange.com/questions/34580/escape-character-in-latex which is from 2011. I'm no expert on LaTeX though.

Next, I would like to ask you to:

1. Make the commit message more concise, like: `fix(latex-renderer): add escaping of % sign`. I'm not sure though whether this is possible to easily do within an existing GitHub PR. So just ignore this if it is not possible.
2. Add some unit test for your change. I think that just extending the existing `test_raw_text()` test method should be enough.

I would merge your PR afterwards. Thank you.

pbodnar:
Ad escaping `_`: It looks like PR #93 (by @sebqq) suggests this, together with not escaping within _inline_ code section... So maybe we should combine it everything together.

jasondenton:
I think that sounds about right. _ needs to be escaped or the results 
are likely to be weird at best and break latex compile at worst.

Inline code blocks are, almost by definition, something that isn't being 
subject to the regular markdown rules. So they should definitely be 
exempt from escaping. I'm not sure what that change looks like in the 
code, so I will leave it to you.

Regards
Jason Denton

Petr Bodnar wrote on 9/16/21 2:24 PM:

pbodnar:
Hi Jason, thanks a lot for your feedback. I will therefore continue with the broader PR #93 and close this one.

pbodnar:
Hi Jason, thanks a lot for your feedback. I will therefore continue with the broader PR #93 and close this one.

Issue #122, opened 2021-10-29
Add missing documentation


pbodnar:
An aggregating documentation and refactoring task, with a dedicated Git branch.

`@mistletoe users`: What are you missing in the docs or in the code, what would you like to improve in this regard?

PR #123, opened 2021-10-29
Add missing documentation (#122)

pbodnar:


pbodnar:
Note: I have focused just on the documentation here in the end.

not-my-profile:
>Important: The parsing phase is currently tightly connected with initiation and closing of a renderer. Therefore, you should never call Document(...) outside of a with ... as renderer block, unless you know what you are doing.

I think it would help to give an example of this "tight connection". Because just by reading that warning I have no idea.

pbodnar:
@not-my-profile, thanks for your feedback -> see the new commit above - I hope it covers the topic of "tight connection" good enough now?

Regarding your second question, I think this should be possible. Something like:

```python
with open('foo.md', 'r') as fin:
    with HTMLRenderer() as renderer:
        doc = Document(fin)
        # ... do some stuff with the doc ...
        rendered = renderer.render(doc)
```

Can you try out and let me know if it works for you?

pbodnar:
OK, so I've pushed the doc commit to master. Feel free to open new issue / PR for improvements. :)

not-my-profile:
Thanks for replying / clarifying!

I guess one of the reasons I find the API confusing is that `Document(..)` doesn't look like it's parsing. I wouldn't expect `Document.__init__` to perform parsing.

pbodnar:
OK, so here we go (8447a42). @not-my-profile, thanks for your feedback! :)

Issue #121, opened 2021-10-27
Make CI work again


pbodnar:
The build on Travis seems to be broken for 2 main reasons now:

1. Python 3.3 seems to be no longer available for download and builds, that's why the last builds kept failing. This version should be therefore possibly removed from ".travis.yml".
2. Builds no longer run on travis-ci.org - migration to travis-ci.com needs to be done ([short setup tutorial](https://docs.travis-ci.com/user/tutorial/)). While builds seem to run there nicely (just testing it now on my forked repo), it looks like they are not completely free and require an extra support request (of an unspecified form) after trial credits are over (see [this post](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing); Travis UI doesn't give too much hints on this so far).

An alternative approach is to use something else than Travis, like GitHub Actions, but I'm not quite familiar with those yet.

pbodnar:
FYI, the integration with Travis CI needs to be setup in "Settings > Integrations > Integrated apps" - only project / account _maintainers_ have got access there though. If we decide to go this way, we will have to ask @miyuchina then...

zyv:
Does it even make sense? Maybe just migrate to built-in GitHub Actions instead?

pbodnar:
> Does it even make sense? Maybe just migrate to built-in GitHub Actions instead?

Sure, GitHub Actions might be also the way to go, as I write:

> An alternative approach is to use something else than Travis, like GitHub Actions, but I'm not quite familiar with those yet.

Tips are welcome. :)

pbodnar:
Getting closer: https://github.com/pbodnar/mistletoe/actions/runs/4107835671 :) The migration to GitHub seems really trivial, just some extra cases (like missing Python 3.6- in the latest Ubuntu) need to be covered. I've also fixed some errors in the legacy code, so that the basic build (linter) won't fail. I guess I will soon push a complete workflow to this repo...

zyv:
> Getting closer: https://github.com/pbodnar/mistletoe/actions/runs/4107835671 :) The migration to GitHub seems really trivial, just some extra cases (like missing Python 3.6- in the latest Ubuntu) need to be covered. I've also fixed some errors in the legacy code, so that the basic build (linter) won't fail. I guess I will soon push a complete workflow to this repo...

Just a tip: you are using an outdated version of `setup-python` - v3 and the current one is v4. This why you are getting deprecation warnings about `set-output` and your runs will start failing soon.

[...]

Issue #114, opened 2021-09-19
LaTeXRenderer: Escape special characters in URLs of generated links


pbodnar:
The LaTeX renderer doesn't currently do any URL escaping in links:

``` python
def render_link(self, token):
    self.packages['hyperref'] = []
    template = '\\href{{{target}}}{{{inner}}}'
    inner = self.render_inner(token)
    return template.format(target=token.target, inner=inner)
```

This may lead to unexpected results / errors. The escaping should be done similarly to what can be seen for example at #102.

pbodnar:
HELP WANTED: I'm no expert on LaTeX and I cannot find a useful documentation on this topic for it quickly. Maybe all that is needed is to URL-encode characters as described at [Encoding of umlaute and special characters in \href links](https://tex.stackexchange.com/questions/417129/encoding-of-umlaute-and-special-characters-in-href-links/417154#417154), i. e. as the other renderers in mistletoe do now?

PR #135, opened 2022-02-13
html.unescape in tokenizer

not-my-profile:
This PR fixes that HTML entities end up in the LaTeX output by the LaTeXRenderer. For details see the individual commits.

If you like these changes, please merge them without squashing.


README.md
```diff
@@ -60,7 +60,7 @@ in the [contrib][contrib] folder:
 
 Installation
 ------------
-mistletoe is tested for Python 3.3 and above. Install mistletoe with pip:
+mistletoe is tested for Python 3.4 and above. Install mistletoe with pip:
```
pbodnar, line 63: OK, looking at <https://w3techs.com/technologies/details/pl-python/3>, I hope their data are correct, v3.3 is only at 0,1%, v3.4 at 2,6%, v3.5 at 8,4% and the mostly used v3.6 at 50%. So we could possibly also drop support for v3.4 as discussed at #134 [here](https://github.com/miyuchina/mistletoe/pull/134#discussion_r805237939) where we talk about requiring v3.5 for `typing`. What do you think?

pbodnar: Also, we shouldn't forget about removing versions from `.travis.yml` as mentioned at #121.

not-my-profile: As far as I am concerned such usage statistics don't matter. What matters is that Python < 3.6 is no longer supported by the Python Software Foundation, see https://endoflife.date/python.

Ok I bumped the version to 3.5 & updated the `.travis.yml`.

Sidenote: The Travis CI does not actually appear to be running? https://app.travis-ci.com/miyuchina/mistletoe

mistletoe/html_renderer.py
```diff
@@ -205,7 +205,7 @@ def render_document(self, token: block_token.Document) -> str:
 
     @staticmethod
     def escape_html(raw: str) -> str:
-        return html.escape(raw).replace('&#x27;', "'")
+        return html.escape(raw)
```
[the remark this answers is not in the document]
not-my-profile: Sure that can be done. Setting `quote=False` for `render_raw_text` does however result in several specification tests failing because the spec tests expect `"` to be escaped as `&quot;` in regular text. We could resolve that by using the official CommonMark `spec_tests.py` test framework (which I am currently also working on rewriting).

Anyway I think that this performance concern should be addressed in a followup PR.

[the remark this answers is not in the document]
pbodnar: OK, let it be as is for now, thank you.

> ... which I am currently also working on rewriting

Looks like you are really into this project now, which is great, but I'm not sure if I will be always able to keep up with your pace. ;)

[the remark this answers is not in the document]
not-my-profile: Ah sorry for the confusion, I meant in the [commonmark-spec](https://github.com/commonmark/commonmark-spec) repository ... not here.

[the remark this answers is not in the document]
pbodnar: Never mind, thanks for the clarification. :)

README.md
pbodnar: OK, thanks.

> Sidenote: The Travis CI does not actually appear to be running? https://app.travis-ci.com/miyuchina/mistletoe

Yes, AFAIK this project is not on the new `.com` instance of Travis yet. Discussion is welcome at #121. :)

mistletoe/html_renderer.py
```diff
@@ -205,11 +205,11 @@ def render_document(self, token: block_token.Document) -> str:
...
-        return html.escape(html.unescape(raw)).replace('&#x27;', "'")
+        return html.escape(raw).replace('&#x27;', "'")
 
     @staticmethod
     def escape_url(raw: str) -> str:
         """
         Escape urls to prevent code injection craziness. (Hopefully.)
         """
-        return html.escape(quote(html.unescape(raw), safe='/#:()*?=%@+,&;'))
+        return html.escape(quote(raw, safe='/#:()*?=%@+,&;'))
```
pbodnar, line 218: OK, so I believe we should remove the surplus call to `unescape()` in the other `escape_url()` definitions too, shouldn't we? (The best would probably be to get rid of the duplicate definitions, but maybe leave that for #114...?)

pbodnar:
Thank you. :)

BTW You may have noticed that I'm a fan of [conventional commits](https://www.conventionalcommits.org/), if you would like to give it a try. But I won't force it on you. ;)

PR #141, opened 2022-04-09
Use CommonMark definition of punctuation charset

jaredliw:
## Something goes wrong...

Given 

```markdown
**你好世界。**Hello world!
```

mistletoe's output is

```html
<p><strong>你好世界。</strong>Hello world!</p>
```

while it should be (according to CommonMark's dingus):

```html
<p>**你好世界。**Hello world!</p>
```

## Why this happens?

Root cause (`core_tokens.py` line 9-11):

```python
punctuation = {'!', '"', '#', '$', '%', '&', '\'', '(', ')', '*', '+', ',',
               '-', '.', '/', ':', ';', '<', '=', '>', '?', '@', '[', '\\',
               ']', '^', '_', '`', '{', '|', '}', '~'}
```

`。` (Chinese fullstop) is not in the set, therefore `**` is regconised as a right-flanking delimiter run (which it shouldn't be).

We should use a broader punctuation charset, including CJK punctuations (and more) as well.

> CommonMark Spec:
> 
> image

**See "Files changed" tab, I left some comments on the modification I made.**

Signed-off-by: jaredliw <jaredliw@gmail.com>

### DIFF
--- a/mistletoe/core_tokens.py
+++ b/mistletoe/core_tokens.py
@@ -1,18 +1,17 @@
 import re
-
+import unicodedata
 
 whitespace = {' ', '\t', '\n', '\x0b', '\x0c', '\r'}
-unicode_whitespace = {'\t', '\n', '\x0b', '\x0c', '\r', '\x1c', '\x1d', '\x1e',
-        '\x1f', ' ', '\x85', '\xa0', '\u1680', '\u2000', '\u2001', '\u2002',
-        '\u2003', '\u2004', '\u2005', '\u2006', '\u2007', '\u2008', '\u2009',
-        '\u200a', '\u2028', '\u2029', '\u202f', '\u205f', '\u3000'}
-punctuation = {'!', '"', '#', '$', '%', '&', '\'', '(', ')', '*', '+', ',',
-               '-', '.', '/', ':', ';', '<', '=', '>', '?', '@', '[', '\\',
-               ']', '^', '_', '`', '{', '|', '}', '~'}
 code_pattern = re.compile(r"(?<!\\|`)(?:\\\\)*(`+)(?!`)(.+?)(?<!`)\1(?!`)", re.DOTALL)
-
-
 _code_matches = []
+
+
+def is_unicode_whitespace(character):
+    return unicodedata.category(character) == "Zs" or character in ("\t", "\u000a", "\u000c", "\000d")
+
+
+def is_unicode_punctuation(character):
+    return unicodedata.category(character) in ("Pc", "Pd", "Pe", "Pf", "Pi", "Po", "Ps")
 
 
 def find_core_tokens(string, root):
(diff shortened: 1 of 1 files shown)


mistletoe/core_tokens.py
```diff
@@ -1,18 +1,17 @@
 import re
-
+import unicodedata
 
 whitespace = {' ', '\t', '\n', '\x0b', '\x0c', '\r'}
-unicode_whitespace = {'\t', '\n', '\x0b', '\x0c', '\r', '\x1c', '\x1d', '\x1e',
-        '\x1f', ' ', '\x85', '\xa0', '\u1680', '\u2000', '\u2001', '\u2002',
-        '\u2003', '\u2004', '\u2005', '\u2006', '\u2007', '\u2008', '\u2009',
-        '\u200a', '\u2028', '\u2029', '\u202f', '\u205f', '\u3000'}
-punctuation = {'!', '"', '#', '$', '%', '&', '\'', '(', ')', '*', '+', ',',
```
jaredliw, line 9: > spec:
>
> A [Unicode punctuation character](https://spec.commonmark.org/0.30/#unicode-punctuation-character) is an [ASCII punctuation character](https://spec.commonmark.org/0.30/#ascii-punctuation-character) or anything in the general Unicode categories Pc, Pd, Pe, Pf, Pi, Po, or Ps.

I didn't add more punctuations to this set due to the following reasons:
1. In the future, there might be new characters in the Unicode categories above. I think it is better for use to use the `unicodedata` library so that we don't change this definition frequently.
2. `unicodedata` doesn't provide a way to get all characters in a specific Unicode category, so, I wrote `is_*` functions instead.
3. Usage of `is_*` functions are more efficient; rather than listing all the things out, we just need to check the given character is a punctuation/whitespace or not.

mistletoe/core_tokens.py
```diff
@@ -1,18 +1,17 @@
 import re
-
+import unicodedata
 
 whitespace = {' ', '\t', '\n', '\x0b', '\x0c', '\r'}
-unicode_whitespace = {'\t', '\n', '\x0b', '\x0c', '\r', '\x1c', '\x1d', '\x1e',
```
jaredliw, line 5: The number of items in this set is different from what is defined in the spec.

Due to sick of checking those non-readable characters one by one, I simply change it to a `is_unicode_whitespace` function.

See the comment on line 9 for more.

mistletoe/core_tokens.py
```diff
@@ -1,18 +1,17 @@
 import re
-
+import unicodedata
 
 whitespace = {' ', '\t', '\n', '\x0b', '\x0c', '\r'}
```
jaredliw, line 4: I would like to arise a question, when do use `whitespace` set instead of `unicode_whitespace` set? Is it possible to use `unicode_whitespace` to replace this set?

I don't get too much into the code, hope this question won't sound stupid xD.

PR #145, opened 2022-06-03
use void tags as per HTML5 spec

elandorr:
A small suggestion to be compliant with https://html.spec.whatwg.org/multipage/syntax.html#void-elements. Void elements don't need the closing slash.

### DIFF
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -63,7 +63,7 @@ class HTMLRenderer(BaseRenderer):
         return template.format(self.render_inner(token))
 
     def render_image(self, token: span_token.Image) -> str:
-        template = '<img src="{}" alt="{}"{} />'
+        template = '<img src="{}" alt="{}"{}>'
         if token.title:
             title = ' title="{}"'.format(html.escape(token.title))
         else:
@@ -188,11 +188,11 @@ class HTMLRenderer(BaseRenderer):
 
     @staticmethod
     def render_thematic_break(token: block_token.ThematicBreak) -> str:
-        return '<hr />'
+        return '<hr>'
 
     @staticmethod
     def render_line_break(token: span_token.LineBreak) -> str:
-        return '\n' if token.soft else '<br />\n'
+        return '\n' if token.soft else '<br>\n'
 
     @staticmethod
     def render_html_block(token: block_token.HTMLBlock) -> str:


pbodnar:
@elandorr, thanks for your PR, but I'm afraid I cannot accept it as-is. The fact is that outputting for example `<br />` **is** already HTML5-compliant, while it also has the practical effect of being XML-compliant. See for example <https://stackoverflow.com/questions/1946426/html-5-is-it-br-br-or-br/1946442#1946442> for a quick recap.

So, I think that if we would also want to support the `<br>` variant, we should state a reason for it (is it required by some wide spread tool, for example?) AND we would have to make this somehow (I've got no concrete idea on this yet) an optional switch in mistletoe.

elandorr:
@pbodnar 

Thanks for the SO link, but the spec is clear, it's just tolerated:

>if the element is one of the void elements, or if the element is a foreign element, then there may be a single U+002F SOLIDUS character (/). **This character has no effect on void elements**

Google does it according to spec:
https://google.github.io/styleguide/htmlcssguide.html#Document_Type

cmark as well:
https://github.com/commonmark/cmark

Not sure how useful XML-compliance as suggested in the SO post is for an explicit HTML renderer. The usual parsers are fine with HTML5. (I have to further edit the output to fix e.g. table aligns as they aren't compliant, and that works fine.)

A switch is always nice; choice.

Don't think any tool cares about it. Real world pages don't follow standards today. It's only a small detail. Not even buzzword bootstrap follows any standard; to use it you have to accept the classic hacks :). Last time someone gave a damn about standards was with xhtml probably, when people proudly put the w3c validator icon up.

Every other part of a current project follows the spec, and having it mixed felt icky. I'm obsessing over details, feel free to ignore. Trying to create something standards-conform that will last til doomsday. Figured I'd attempt to follow as close as possible. (Semantics are still very ill-defined for example, not even the w3c itself follows their standard!)

Have a great remaining Sunday :)

pbodnar:
@elandorr, thanks for your feedback, maybe you won't like me, but I believe I'm as pragmatic as can be in this area:

> Thanks for the SO link, but the spec is clear, it's just tolerated:
> 
> > if the element is one of the void elements, or if the element is a foreign element, then there may be a single U+002F SOLIDUS character (/). **This character has no effect on void elements**

Well, when something is optional, it doesn't mean it is not widely used or is planned to be deprecated, right? AFAIK HTML5 was intentionally created with the possibility to produce well formed XML documents.

> Google does it according to spec: https://google.github.io/styleguide/htmlcssguide.html#Document_Type

Well, this seems to be a style-guide for a large, yet limited group of people working on Google stuff. I would be cautious to take this as a "reference" for how to write HTML (see [this opinion](http://xahlee.info/js/Google_html_css_style_guide.html), for example).

> cmark as well: https://github.com/commonmark/cmark

Commonmark clearly states in their [tests](https://github.com/commonmark/cmark/blob/master/test/spec.txt#L9216) that `<br />` should be output. But maybe I have missed something?
 
> Not sure how useful XML-compliance as suggested in the SO post is for an explicit HTML renderer. The usual parsers are fine with HTML5. (I have to further edit the output to fix e.g. table aligns as they aren't compliant, and that works fine.)
>
> A switch is always nice; choice.

Sure, but we are talking about that someone might rely on a tool which further processes the HTML output from mistletoe and feeding the output directly to an XSLT processor, for example. **And if the processor doesn't know anything about HTML "gotchas" like void tags without a slash, we would break their function.**

> Don't think any tool cares about it. Real world pages don't follow standards today. It's only a small detail. Not even buzzword bootstrap follows any standard; to use it you have to accept the classic hacks :). Last time someone gave a damn about standards was with xhtml probably, when people proudly put the w3c validator icon up.

Yes, standards come and go, real-world scenarios & making our life simpler by selecting "the best possible intersection" should be our goal. For me, this is the slash variant for void tags (also required by SVG, BTW).
 
> Every other part of a current project follows the spec, and having it mixed felt icky. I'm obsessing over details, feel free to ignore. Trying to create something standards-conform that will last til doomsday. Figured I'd attempt to follow as close as possible. (Semantics are still very ill-defined for example, not even the w3c itself follows their standard!)

I hope it is clear now that we are following the standards here and this should not cease to be truth sometime in the foreseeable future. OTOH, I'm not strictly against giving users also the choice to use the no-slash variant, provided there will be enough reasoning & interest in that.

> Have a great remaining Sunday :)

Thanks, you too. :)

pbodnar:
@elandorr, it looks like you really enjoy to discuss stuff, don't you? :) Frankly, I don't get all the

PR #151, opened 2022-07-07
Remove reference to dev branch in CONTRIBUTING.md

joel-coffman:
The dev branch hasn't been updated in more than 3 years. Thus, the
recommendation to "check...the dev branch regularly" because "the
master branch is updated only when features on the dev branch are
stabilized" is misleading.


joel-coffman:
Tripped me up when I was preparing my change (#150) so I figured I might as well update the documentation.

pbodnar:
@joel-coffman, good point, thank you. I think the dev branch really is an overkill for us, at least at this stage of the project and when nearly everything is always being developed and tested in a dedicated branch firstly.

Issue #142, opened 2022-04-09
0.8.2: pytest is failing in `test/test_contrib/test_github_wiki.py::TestGithubWiki::test_parse` unit and `DeprecationWarning` warnings


kloczek:
I'm trying to package your module as an rpm package. So I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix>
- run pytest with PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>

Here is pytest output:
```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-mistletoe-0.8.2-2.fc35.x86_64/usr/lib64/python3.8/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-mistletoe-0.8.2-2.fc35.x86_64/usr/lib/python3.8/site-packages
+ /usr/bin/pytest -ra
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.8.13, pytest-7.1.1, pluggy-1.0.0
rootdir: /home/tkloczko/rpmbuild/BUILD/mistletoe-0.8.2
collected 208 items

test/test_ast_renderer.py ..                                                                                                                                         [  0%]
test/test_block_token.py ..................................s...........                                                                                              [ 23%]
test/test_cli.py ...............                                                                                                                                     [ 30%]
test/test_core_tokens.py ...........                                                                                                                                 [ 35%]
test/test_html_renderer.py .............................                                                                                                             [ 49%]
test/test_latex_renderer.py ........................                                                                                                                 [ 61%]
test/test_latex_token.py .                                                                                                                                           [ 61%]
test/test_span_token.py ....................                                                                                                                         [ 71%]
test/test_traverse.py .                                                                                                                                              [ 71%]
test/test_contrib/test_github_wiki.py F.                                                                                                                             [ 72%]
test/test_contrib/test_jira_renderer.py ..................................                                                                                           [ 88%]
test/test_contrib/test_mathjax.py ..                                                                                                                                 [ 89%]
test/test_contrib/test_toc_renderer.py ......                                                                                                                        [ 92%]
test/test_contrib/test_xwiki20_renderer.py ...............                                                                                                           [100%]

================================================================================= FAILURES =================================================================================
[...]

PR #153, opened 2022-08-09
Bug fix for #142: pytest failing.

anderskaplan:
Hi, here's a fix for issue #142. The problem was that the `GithubWiki` token didn't get added to `span_token._token_types` as intended under pytest.

### DIFF
--- a/test/test_contrib/test_github_wiki.py
+++ b/test/test_contrib/test_github_wiki.py
@@ -1,6 +1,6 @@
 from unittest import TestCase, mock
 from mistletoe import span_token, Document
-from mistletoe.span_token import tokenize_inner, _token_types
+from mistletoe.span_token import tokenize_inner
 from contrib.github_wiki import GithubWiki, GithubWikiRenderer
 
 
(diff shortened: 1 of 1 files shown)


pbodnar:
@anderskaplan, thanks for the fix, I will hopefully check it out soon.

As you haven't quite explained what was the culprit, I'm curious if your removal of import of `span_token._token_types` plays a role in the fix, or not...

anderskaplan:
It does. I'm not quite sure what the cause of the problem is, but I'd guess it has something to do with how modules are loaded and module-global data is initialized when running within pytest.

So the solution was to reference `_token_types` with module specifier in the unit test. And then there was no need for the import.

pbodnar:
@anderskaplan, thanks for the explanation. 👍  I've just checked myself and have come to the same conclusion.

So the second change of removing `autospec` argument from the call to `mock.Mock()` is apparently just a cosmetic one - it looks like the argument was useless.

So I will leave this PR open for a day or two for the case you would like to put the explanation right into the commit message (which I believe is a best practice) before I merge it... :)

anderskaplan:
Ok, updated the commit message as requested.

And you're right, the Mock `autospec` argument isn't even implemented for the Mock constructor, so removing it was a cosmetic thing.

pbodnar:
@anderskaplan, I'm sorry for reopening the topic, but based on the previous comments, I thought rather something like this:

```
Bug fix for #142: pytest failing.

The problem was that when run under pytest, the test referenced a wrong instance of `_token_types`,
therefore setting mocked RawText to a wrong list.
Solved by referencing `_token_types` with module specifier in the unit test.
```

So can you please amend the commit once again? I promise it will be for the last time. ;)

anderskaplan:
No problem. I've updated the commit message.

pbodnar:
Thank you. :)

Issue #108, opened 2021-06-23
Update to CommonMark v0.30


enderger:
Recently, the CommonMark spec got an update. While the project may be inactive, if an update was awaiting the spec now is the time to do it.

pbodnar:
OK, I've just checked against v0.30 and 20 tests of 652 are failing. It will be quite a job to get all fixed, I guess. Any volunteers? :)

This is how to download new spec and run tests against it:

1) Change `VERSION` in "test/specification/spec.sh".
2) Run in bash:

```
$ cd test/specification
$ ./spec.sh
...
$ cd ../..
$ python -m test.specification
```

pbodnar:
About to be fixed by: #156.

anderskaplan:
Example 28 fails because HTML5 entities without a trailing semicolon (such as &copy) are explicitly not allowed by the CommonMark spec, but mistletoe uses the python function `html.unescape()` under the hood, which *does* recognize them.

This is a case where it can be argued whether spec compliance is an improvement, or not. Thoughts?

I certainly see the value in spec compliance, but I'll wait with fixing this anyway to give some time for discussion.

pbodnar:
> Example 28 fails because HTML5 entities without a trailing semicolon (such as &copy) are explicitly not allowed by the CommonMark spec, but mistletoe uses the python function `html.unescape()` under the hood, which _does_ recognize them.
> 
> This is a case where it can be argued whether spec compliance is an improvement, or not. Thoughts?
> 
> I certainly see the value in spec compliance, but I'll wait with fixing this anyway to give some time for discussion.

Well, I have tested and it seems like in fact mistletoe is already pretty close to spec compliance - this is the output from the failed test:

```
example:  28
markdown: '&nbsp &x; &#; &#x;\n&#87654321;\n&#abcdef0;\n&ThisIsNotDefined; &hi?;\n'
html:     '<p>&amp;nbsp &amp;x; &amp;#; &amp;#x;\n&amp;#87654321;\n&amp;#abcdef0;\n&amp;ThisIsNotDefined; &amp;hi?;</p>\n'
output:   '<p>&amp;nbsp &amp;x; &amp;#; &amp;#x;\n�\n&amp;#abcdef0;\n&amp;ThisIsNotDefined; &amp;hi?;</p>\n'
```

[...]

PR #156, opened 2022-08-20
Fix for part of #108, Update to CommonMark v0.30

anderskaplan:
This PR fixes nine of the failing examples in the CommonMark 0.30 specification. They all had in common that content inside code spans was not handled according to the spec.

This was solved by preserving space and escape sequences during parsing, and by removing leading and trailing space according to the spec during HTML rendering.

The PR also includes a fix for a warning, a simplified way to download the spec tests examples, and an improvement to the spec test runner.

### DIFF
--- a/test/test_span_token.py
+++ b/test/test_span_token.py
@@ -53,6 +53,14 @@ class TestInlineCode(TestBranchToken):
     def test_parse_in_strikethrough(self):
         self._test_parse_enclosed(span_token.Strikethrough, '~~')
 
+    def test_preserve_space(self):
+        self._test_parse(span_token.InlineCode, '``` ```', ' ')
+        self._test_parse(span_token.InlineCode, '`  ``  `', '  ``  ')
+
+    def test_preserve_escapes(self):
+        self._test_parse(span_token.InlineCode, '`\\xa0b\\xa0`', '\\xa0b\\xa0')
+        self._test_parse(span_token.InlineCode, '`` \\[\\` ``', ' \\[\\` ')
+
 
 class TestStrikethrough(TestBranchToken):
     def test_parse(self):
--- a/test/test_span_token.py
+++ b/test/test_span_token.py
@@ -98,10 +98,10 @@ class TestImage(TestBranchToken):
 
 class TestEscapeSequence(TestBranchToken):
     def test_parse(self):
-        self._test_parse(span_token.EscapeSequence, '\*', '*')
+        self._test_parse(span_token.EscapeSequence, r'\*', '*')
 
     def test_parse_in_text(self):
-        tokens = iter(span_token.tokenize_inner('some \*text*'))
+        tokens = iter(span_token.tokenize_inner(r'some \*text*'))
         self._test_token(next(tokens), 'some ', children=False)
         self._test_token(next(tokens), '*')
         self._test_token(next(tokens), 'text*', children=False)
--- a/test/test_span_token.py
+++ b/test/test_span_token.py
@@ -53,13 +53,13 @@ class TestInlineCode(TestBranchToken):
     def test_parse_in_strikethrough(self):
         self._test_parse_enclosed(span_token.Strikethrough, '~~')
 
-    def test_preserve_space(self):
+    def test_remove_space_if_present_on_both_sides(self):
         self._test_parse(span_token.InlineCode, '``` ```', ' ')
-        self._test_parse(span_token.InlineCode, '`  ``  `', '  ``  ')
+        self._test_parse(span_token.InlineCode, '`  ``  `', ' `` ')
 
     def test_preserve_escapes(self):
         self._test_parse(span_token.InlineCode, '`\\xa0b\\xa0`', '\\xa0b\\xa0')
-        self._test_parse(span_token.InlineCode, '`` \\[\\` ``', ' \\[\\` ')
+        self._test_parse(span_token.InlineCode, '``\\`\\[``', '\\`\\[')
 
 
 class TestStrikethrough(TestBranchToken):
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -54,7 +54,10 @@ class HTMLRenderer(BaseRenderer):
 
     def render_inline_code(self, token: span_token.InlineCode) -> str:
         template = '<code>{}</code>'
-        inner = html.escape(token.children[0].content)
+        content = token.children[0].content.replace('\n', ' ')
+        if not content.isspace() and content.startswith(" ") and content.endswith(" "):
+            content = content[1:-1]
+        inner = html.escape(content)
         return template.format(inner)
 
     def render_strikethrough(self, token: span_token.Strikethrough) -> str:
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -54,10 +54,7 @@ class HTMLRenderer(BaseRenderer):
 
     def render_inline_code(self, token: span_token.InlineCode) -> str:
         template = '<code>{}</code>'
-        content = token.children[0].content.replace('\n', ' ')
-        if not content.isspace() and content.startswith(" ") and content.endswith(" "):
-            content = content[1:-1]
-        inner = html.escape(content)
+        inner = html.escape(token.children[0].content)
         return template.format(inner)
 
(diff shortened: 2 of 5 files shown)


pbodnar:
@anderskaplan, thanks for another contribution, I will try to look at this soon. Where do _you_ take time for it all? ;)

BTW I think that with all the past and most recent fixes, we can slowly release version 1.0.0 of mistletoe...? Also note the project's status equal to "alpha" on <https://pypi.org/project/mistletoe/> - not sure if this isn't unnecessarily low, it looks like this can be changed in [setup.py](https://github.com/miyuchina/mistletoe/blob/master/setup.py), maybe I should create a dedicated issue for that... :)

anderskaplan:
You're welcome! 😄 It's a nice hobby project.

I don't have much of an opinion about when and how to release the 1.0.0, but I agree that "alpha" sounds unnecessarily low. Especially given that there are quite a few projects depending on it already!

pbodnar:
> You're welcome! 😄 It's a nice hobby project.

Same here. :) And I'm primarily using it via CopyQ to copy-paste my markdown notes to whatever format I need... :)

> 
> I don't have much of an opinion about when and how to release the 1.0.0, but I agree that "alpha" sounds unnecessarily low. Especially given that there are quite a few projects depending on it already!

OK, thanks for your feedback, then I would at minimum remove the "alpha" from the mistletoe version soon. :)

pbodnar (changes requested):
To the PR, thanks again, I find it nearly perfect,

1) I would suggest just 1 change: as the updated rules for white(space) at <https://spec.commonmark.org/0.30/#code-span> (for the record, these updates come from version [0.29](https://raw.githubusercontent.com/commonmark/commonmark-spec/0.29.0/changelog.txt)) seem format-independent to me, I would still keep this logic inside `span_token.py` rather than in `html_renderer.py` - so can you please move the new code there?

2) I was a bit confused about the CommonMark's 0.28 example no. 318 `foo...bar\n..baz` (`.` = ` `) being rendered as a code containing `foo...bar.baz`, but it seems OK, because GitHub renderer seems to do the same when it finds _spaces at the beginning of a line_. Also, this example was changed to `\nfoo\nbar..\nbaz\n`, which is correctly treated as `foo.bar...baz`. So no problem, just for the record here.

3)

> This PR fixes **ten** of the failing examples in the CommonMark 0.30 specification. ...

If I count correctly, it is only **nine** out of 20 failed examples, but still great! :) And do I guess it right that you are about to look at the rest of failing examples as well, or maybe you have already done so? :)

4) I'm not quite sure about the 3rd commit (a53878d5db748b6dca438dcbf6d35e7fcb5cd998) - maybe for a separate issue / PR? It looks like something I have coped with here: https://github.com/miyuchina/mistletoe/commit/0daa6058db816a1c7a6391f03185f764210f5003#diff-fb53ae72545ff8771393e790376f6589bfc81a0909c879c82929b05dc57d903d - maybe a similar approach would also fix the problem with `stdout` on Windows? I'm on Windows, using Git bash, and I haven't noticed the problem when running the specification tests yet, but maybe I have done some tuning in the past, I don't know now...

test/test_span_token.py
```diff
@@ -53,6 +53,14 @@ def test_parse_in_emphasis(self):
     def test_parse_in_strikethrough(self):
         self._test_parse_enclosed(span_token.Strikethrough, '~~')
 
+    def test_preserve_space(self):
+        self._test_parse(span_token.InlineCode, '``` ```', ' ')
+        self._test_parse(span_token.InlineCode, '`  ``  `', '  ``  ')
+
+    def test_preserve_escapes(self):
+        self._test_parse(span_token.InlineCode, '`\\xa0b\\xa0`', '\\xa0b\\xa0')
```
pbodnar, line 61: > Solved by converting line endings to space and preserving space and escape sequences in InlineCode tokens

OK, so this one tests that escape sequences are preserved - but do I see it correctly, that regarding _escape sequences_, no change was necessary / done?

anderskaplan: That's right, there is a new test case to check that escape sequences are preserved according to the spec, but no change to the logic was required.

anderskaplan: Great, thanks for the thorough review!

1. I understand your point of wanting to make minimal changes and keep logic where it is. However, in this case, I'd like to ask you to reconsider -- let me explain why! Currently, the token classes do the parsing and some processing, and the renderer classes process the input, too. It would be helpful if the token classes would limit themselves to only do the parsing, and leave all processing to the renderers -- because different renderers may have different needs. I'm working on a Markdown renderer, for example, where the goal is to be able to do a near-perfect round trip. That's not possible if the source has been processed in an irreversible way. (I am nearly finished with a first iteration of that renderer, and it does have a few more little source-preserving changes like this one.) I am of course fully aware that mistletoe was not written with "my" use case in mind, but I do believe that a clear separation of concerns between parsing and rendering can benefit the project as a whole anyway.

3. Correct. The classic off-by-one.

4. If you try and run the 0.30 tests on Windows, you'll probably see the problem. One of the new, failing test cases emits an invalid Unicode codepoint and then the test fails with an exception because of its own printout. It's for sure not an essential patch; I just like tests to run cleanly.

pbodnar: Thanks for your feedback, here are my remarks to the 1st point:

Generally, I agree with you, parsing and rendering should be as separated as possible. In this concrete case though, I still believe that this new handling of extra spaces around the inline code itself should be considered as related just to parsing.

Yes, every single renderer could decide itself what to do with those extra spaces, but I don't see much added value here. I think it would be similar to leaving renderers decide whether they want to unescape entity references, or not (unescaping is done on the parsing level since #135).

Of course, the renderer _to Markdown_ which you prepare would most probably benefit from knowing that extra spaces were used, so it can preserve the original spacing. Yet isn't this a special case of a renderer? I expect that you already add some extra info to the parsed tokens, so I can imagine that storing an additional flag of "extra spaces present" could do the job?

pbodnar: Ad 3. I've checked again and I still can't see the "invalid Unicode codepoint" problem - just 11 failed spec tests (running `py -m test.specification` with v0.30 of CommonMark downloaded). So probably yes, for a better focus on just that problem, I would suggest to move it into an extra PR. Thanks in advance.

PS: I've just noticed the last 3 comments should be moved 1 level up (out of code line conversation), but whatever, next time... :)

pbodnar: Ad 1. There is also this stuff of _replacing new lines with space_ within inline code - I think that this is also something that authors of the CommonMark spec consider "getting rid of input formatting which is not meant to be presented by renderers". I hope it makes sense now why I'm against keeping this logic on the renderers' part, I hope it wasn't too verbose... :)

anderskaplan:
There, updated according to comments.

test/test_span_token.py
anderskaplan: Sure, though I doubt that the authors of the CommonMark spec care much about other renderers/transformations than Markdown to HTML. 😃

pbodnar: OK, thanks for the amendments.

Sure, the primary output format is definitely HTML, so it is only logical that the spec authors illustrate the parsing results on the rendered HTML. It's hard for me to tell how much they actually care about the other output formats, but I feel the spec is written in a way that it basically focuses on how to interpret the input markdown, getting an AST as a result, and that's it... :)

pbodnar:
@anderskaplan, thank you. So if you're not against, I'm going to put the commits as-is onto the master branch - even though changes in the last commit could be probably squashed into its parent commit, I think it is not that critical to do...

pbodnar (approved):

test/test_span_token.py
anderskaplan: I can just add that my understanding is that the spec only describes how to get from Markdown to HTML. The AST is an implementation detail not covered by the spec, where Markdown-to-HTML processors have full artistic freedom. But I may be wrong. And of course, AST users are probably not too happy about breaking changes to the API.

pbodnar: Yeah, it looks we see it a little bit differently, partly meaning the same thing, but using different words for it (like when I agree with you a concrete AST really is an implementation detail of a given parser), but I think that never mind. :) BTW Good point about a possible "breaking change", I will try to mention it in the next release notes somehow...

PR #158, opened 2022-09-02
fix: Apply traverse class filter to source when include_source=True

asb:
Previously, any class filter given to `traverse` wouldn't be respected
for the source node (returned if `include_source=True`). This patch
addresses this.

This is considered a bug fix as opposed to a breaking change because:
* Filtering by class was broken anyway in the last mistletoe release.
* It is unintuitive and likely unintended for the filter not to apply to
  the source node.


pbodnar:
@asb, a pro work, thank you :)

PR #159, opened 2022-09-02
fix: Make limiting depth in traverse() actually work

asb:
Due to a typo (inverted comparison), the test for the traversal depth would always fail. Fix that typo and add test coverage.

### DIFF
--- a/test/test_traverse.py
+++ b/test/test_traverse.py
@@ -70,3 +70,56 @@ class TestTraverse(unittest.TestCase):
         ]
         self.assertEqual(filtered, ["Strong", "Strong"])
 
+    def test_with_depth_limit(self):
+        doc = Document("a **b** c **d**")
+
+        # Zero depth with root not included yields no nodes.
+        filtered = [t.node.__class__.__name__ for t in traverse(doc, depth=0)]
+        self.assertEqual(filtered, [])
+
+        # Zero depth with root included yields the root node.
+        filtered = [
+            t.node.__class__.__name__
+            for t in traverse(doc, depth=0, include_source=True)
+        ]
+        self.assertEqual(filtered, ["Document"])
+
+        # Depth=1 correctly returns the single node at that level.
+        filtered = [t.node.__class__.__name__ for t in traverse(doc, depth=1)]
+        self.assertEqual(filtered, ["Paragraph"])
+
+        # Depth=2 returns the correct nodes.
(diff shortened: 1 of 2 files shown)


test/test_traverse.py
```diff
@@ -70,3 +70,44 @@ def test_with_included_source_and_class_filter(self):
         ]
         self.assertEqual(filtered, ["Strong", "Strong"])
 
+    def test_with_depth_limit(self):
+        doc = Document("a **b** c **d**")
+        filtered = [t.node.__class__.__name__ for t in traverse(doc, depth=0)]
```
pbodnar, line 75: It seems like all scenarios are covered -> OK.

Just for a little bit better readability, I would suggest to separate them by a new line and to add a short description on top of each - e.g. `# zero depth with root not included yields no nodes` for the 1st one (of course, we could put a message to the `assertEqual()` call as well, but I think it is not necessary in this case).

pbodnar (changes requested):
@asb, great, just 1 little suggestion.

I have just merged the previous PR, so you can also do Git rebase right away.

test/test_traverse.py
asb: Thanks for the feedback - there's a range of views on when and if to split into multiple test functions, and how to comment tests so I expected feedback in this area.

Separating the tests visually and adding a comment seems very pragmatic to me - I've made that change and pushed.

pbodnar: Thank you. :)

test/test_traverse.py
```diff
@@ -70,3 +70,56 @@ def test_with_included_source_and_class_filter(self):
...
+            t.node.__class__.__name__
+            for t in traverse(doc, depth=0, include_source=True)
+        ]
+        self.assertEqual(filtered, ["Document"])
+
+        # Depth=1 correctly returns the single node at that level.
+        filtered = [t.node.__class__.__name__ for t in traverse(doc, depth=1)]
+        self.assertEqual(filtered, ["Paragraph"])
+
+        # Depth=2 returns the correct node.
```
pbodnar, line 91: I think it's exactly as I have wished, probably just a small typo here?

```suggestion
        # Depth=2 returns the correct nodes.
```

asb:
Thanks, typo fixed.

pbodnar:
Merged, thank you!

PR #161, opened 2022-09-09
💡 implement link titles in the JIRARenderer (contrib)

franferrax:
Sample _Markdown_:
```md
Sample: [inner](https://target "title")
```

In _Linux_, test with:
```sh
mistletoe -r contrib.jira_renderer.JIRARenderer <(echo 'Sample: [inner](https://target "title")')
```

_Jira_ result, before this PR changes:
```
Sample: [inner|https://target]
```

_Jira_ result, after this PR changes:
```
Sample: [inner|https://target|title]
```

### DIFF
--- a/contrib/jira_renderer.py
+++ b/contrib/jira_renderer.py
@@ -63,10 +63,15 @@ class JIRARenderer(BaseRenderer):
         return template.format(src=token.src)
 
     def render_link(self, token):
-        template = '[{inner}|{target}]'
-        target = escape_url(token.target)
+        template = '[{inner}|{target}{title}]'
         inner = self.render_inner(token)
-        return template.format(target=target, inner=inner)
+        target = escape_url(token.target)
+        if token.title:
+            title = '|{}'.format(token.title)
+        else:
+            title = ''
+
+        return template.format(inner=inner, target=target, title=title)
 
     def render_auto_link(self, token):
         template = '[{target}]'
--- a/test/test_contrib/test_jira_renderer.py
+++ b/test/test_contrib/test_jira_renderer.py
@@ -102,6 +102,15 @@ class TestJIRARenderer(BaseRendererTest):
         expected = '[{body}|{url}]'.format(url=url, body=body)
         actual = self.renderer.render(token)
         self.assertEqual(expected, actual)
+
+    def test_render_link_with_title(self):
+        url = 'http://{0}.{1}.{2}'.format(self.genRandomString(5), self.genRandomString(5), self.genRandomString(3))
+        body = self.genRandomString(80, True)
+        title = self.genRandomString(20, True)
+        token = next(iter(tokenize_inner('[{body}]({url} "{title}")'.format(url=url, body=body, title=title))))
+        expected = '[{body}|{url}|{title}]'.format(url=url, body=body, title=title)
+        actual = self.renderer.render(token)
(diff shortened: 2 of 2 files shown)


contrib/jira_renderer.py
```diff
@@ -63,10 +63,14 @@ def render_image(self, token):
         return template.format(src=token.src)
 
     def render_link(self, token):
-        template = '[{inner}|{target}]'
```
pbodnar, line 66: OK, this seems to be in-line with the existing code here.

Yet, while being at it, what about also swapping the order of creating and passing values to the `template`, so that it is more natural (like inside the `render_block_code()` method)?

pbodnar (commented):
@franferrax, great, thanks for your contribution.

Do I see it right, that while this works, it doesn't seem to be [documented](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=links) (or anywhere else) by Atlassian?

Next, do you think you could also add a unit test to cover this new feature? I think cloning and extending the existing `test_render_link()` as `test_render_link_with_title()` could be sufficient. :)

contrib/jira_renderer.py
franferrax: @pbodnar: sure, no problem, I was just trying to follow the style and minimize the diff, I'll amend and force-push that one.

franferrax:
> @franferrax, great, thanks for your contribution.
> 
> Do I see it right, that while this works, it doesn't seem to be [documented](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=links) (or anywhere else) by Atlassian?
> 
Oh you are right! I know that documentation, I just discovered this by accident and hadn't realized this isn't there. Feel free to discard this PR if you prefer to only implement well documented features.

> Next, do you think you could also add a unit test to cover this new feature? I think cloning and extending the existing `test_render_link()` as `test_render_link_with_title()` could be sufficient. :)

Yes, I'll try to do that in a separated commit.

franferrax:
@pbodnar: this PR is now ready for re-review, I've amended and force-pushed 369a5f52459b7039c17ffc671fd04cb3b53726ce &rarr; d0cb67e697276b668995c866017edf1ed22a032e to include [commented](#discussion_r970050994) suggestions. I've added `test_render_link_with_title()`, checked it is passing, and temporarily removed my changes to check it fails.

contrib/jira_renderer.py
pbodnar: I see, thanks for the amendment.

pbodnar:
@franferrax, thanks for having quickly implemented the requested changes. :)

> > @franferrax, great, thanks for your contribution.
> > Do I see it right, that while this works, it doesn't seem to be [documented](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=links) (or anywhere else) by Atlassian?
> 
> Oh you are right! I know that documentation, I just discovered this by accident and hadn't realized this isn't there. Feel free to discard this PR if you prefer to only implement well documented features.
> 

I tried to Google again and I still couldn't find anything on this Jira markup feature, like since which version of JIra it is available. Yet, I think this PR makes sense and it is probably safe to be merged.

So only just for sure, for the case someone follows us, I would leave it still open for a while (like a week or two), giving them a chance to challenge this change. :))

pbodnar (approved):

Issue #101, opened 2020-07-05
Make renderers from the "contrib" folder easier to use


pbodnar:
From my point of view, there are currently two gotchas about the "contrib" folder:

1. If one just installs mistletoe by running `pip install mistletoe` (or `pip install .` after Git cloning), the renderers from the "contrib" folder are not available when running mistletoe from any location. Unless one clones the Git repo and passes full (relative) path to a renderer. (I tried to describe this in [this commit](https://github.com/pbodnar/mistletoe/commit/5993e88a7fab52ada80ace002f27e6f5f87feb40).)
2. The purpose of the whole "contrib" folder is kinda fuzzy. Apart from the various renderers that I suspect are probably a "contributed" unfinished work (see for example #100), there is also for example `md2jira.py` - how does this script fit to the concept of having the main mistletoe cli script that can do all the necessary stuff?

Here are some suggestions:

1. Would it be possible to document the purpose of the "contrib" folder and how to work with it?
2. What about moving the "contrib" folder, or at least renderers from there, to the "mistletoe" folder (where the other "core" renderers reside), or to make a subfolder within it, so that the renderers get available when mistletoe gets installed (not only when "installed" for developing purposes, i. e. by running `pip install -e .`)? Or maybe to have a special module "mistletoe_contrib" next to the existing "mistletoe" module (in the project's root folder) and add some installation scripts for it?

Note: I'm relatively new to Python and this "module / package bundling", so it is quite possible I have overlooked some easier way, I will be glad if someone shows me...

pbodnar:
For the beginning, I have at least tried to document this: 86fe3a87b4d80d596b773c9e50aa6c2b5c4aa826.

anderskaplan:
Hey @pbodnar, I found [this article](https://softwareengineering.stackexchange.com/questions/252053/whats-in-the-contrib-folder) which explains pretty nicely how some popular open source software packages use the contrib folder.

I think your suggestion to move the contrib folder to become a subfolder within the mistletoe folder is good. That's how Django does it. The big advantage, as I see it, is that it makes the tools in contrib available as part of the regular pip package.

I can put up a PR for this if you like.

pbodnar:
@anderskaplan, thanks for your feedback on this (the only feedback so far actually ;)). If you have time, it would be great if you prepare a PR. :)

anderskaplan:
PR created

pbodnar:
And the PR resolved it. :)

PR #162, opened 2022-09-17
Markdown renderer

anderskaplan:
This is a pull request for a Markdown-to-Markdown renderer (#4).

A few notes on the implementation:

1. Since the aim is to preserve as much of the formatting as possible, and also steer clear of issues like if
   the sequence `_**Hello**_` would be rendered as `***Hello***` (which changes its interpretation),
   I had to save some formatting information from the parsing in the tokens. The formatting info will not appear in the output from the ASTRenderer or from repr().

2. The Markdown renderer can also do word wrapping. (Because I needed that for a use case.) But that's optional, and when it's not enabled, the goal is still to make the roundtrip Markdown -> parse -> render -> Markdown with as few changes as possible.

3. The test cases in `TestMarkdownRenderer` give a pretty good idea of what the renderer is capable of.
   In particular, note that it handles nested blocks: you can put a code block inside a list inside a block quote 
   if you like. (This is explicitly allowed by the CommonMark spec.)

4. Three new tokens! Well actually, only one which is completely new and that's the `BlankLine`.
   The other two represent link reference definitions (a.k.a. footnotes) and inherit from the`Footnote` token.
   These tokens are used to represent content that is otherwise not included in the AST.

   An open question is whether they belong in the same module as the markdown renderer class, or if they should
   be moved somewhere else. Maybe create a new module to host all optional tokens?

5. The markdown renderer handles span tokens and block tokens differently.
   Block tokens are rendered into chunks of lines.
   Span tokens are first rendered into sequences of "fragments", which are essentially strings with some additional metadata attached, and then into chunks of lines.

6. Tables are supported.

7. I put the MarkdownRenderer in the main package, not in contrib, because that's what @miyuchina wrote in a comment on #4.

Hope this can be useful and looking forward to some review comments. As it's a fairly large PR, I'd recommend to review it commit by commit. All changes to existing code come first, and then finally the renderer itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ main package:
 * HTML
 * LaTeX
 * AST (Abstract Syntax Tree; handy for debugging the parsing process)
+* Markdown (Can be used to reflow the text, or make other types of automated
+  changes to Markdown documents)
 
 Renderers for the following output formats can be found
 in the [contrib][contrib] package:
@@ -106,6 +108,17 @@ from mistletoe.latex_renderer import LaTeXRenderer
 
 with open('foo.md', 'r') as fin:
     rendered = mistletoe.markdown(fin, LaTeXRenderer)
+```
+
+To reflow the text in a Markdown document with a max line length of 20 characters:
+
+```python
+import mistletoe
+from mistletoe.markdown_renderer import MarkdownRenderer
+
+with open('dev-guide.md', 'r') as fin:
+    with MarkdownRenderer(max_line_length=20) as renderer:
+        print(renderer.render(mistletoe.Document(fin)))
 ```
 
 Finally, here's how you would manually specify extra tokens via a renderer.
--- a/dev-guide.md
+++ b/dev-guide.md
@@ -244,3 +244,59 @@ with open('foo.md', 'r') as fin:
 For more info, take a look at the `base_renderer` module in mistletoe.
 The docstrings might give you a more granular idea of customizing mistletoe
 to your needs.
+
+## Markdown to Markdown
+
+Suppose you have some Markdown that you want to process and then output
+as Markdown again. Thanks to the text-like nature of Markdown, it is often
+possible to do this with text search-and-replace tools... but not always. For
+example, if you want to replace a text fragment in the plain text, but not
+in the embedded code samples, then the search-and-replace approach won't work.
+
+In this case you can use mistletoe's `MarkdownRenderer`:
+1. Parse Markdown to an AST tree (usually held in a `Document` token).
+2. Make modifications to the AST tree.
+3. Render back to Markdown using `MarkdownRenderer.render()`.
+
+Here is an example of how you can make text replacements in selected parts
+of the AST:
+
+```python
+import mistletoe
+from mistletoe.block_token import BlockToken, Heading, Paragraph, SetextHeading
+from mistletoe.markdown_renderer import MarkdownRenderer
+from mistletoe.span_token import InlineCode, RawText, SpanToken
+
+def update_text(token: SpanToken):
+    """Update the text contents of a span token and its children.
+    `InlineCode` tokens are left unchanged."""
+    if isinstance(token, RawText):
+        token.content = token.content.replace("mistletoe", "The Amazing mistletoe")
+
+    if not isinstance(token, InlineCode) and hasattr(token, "children"):
+        for child in token.children:
+            update_text(child)
+
+def update_block(token: BlockToken):
+    """Update the text contents of paragraphs and headings within this block,
+    and recursively within its children."""
+    if isinstance(token, (Paragraph, SetextHeading, Heading)):
+        for child in token.children:
+            update_text(child)
+
+    for child in token.children:
+        if isinstance(child, BlockToken):
+            update_block(child)
+
+with open("README.md", "r") as fin:
+    with MarkdownRenderer() as renderer:
+        document = mistletoe.Document(fin)
+        update_block(document)
+        md = renderer.render(document)
+        print(md)
+```
+
+If you're making large changes, so that the formatting of the document is
+affected, then it can be useful to also have the text reflowed. This can
+be done by specifying a `max_line_length` parameter in the call to the
+`MarkdownRenderer` constructor.
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -376,14 +376,20 @@ class BlockCode(BlockToken):
     @classmethod
     def read(cls, lines):
         line_buffer = []
+        trailing_blanks = 0
         for line in lines:
             if line.strip() == '':
                 line_buffer.append(line.lstrip(' ') if len(line) < 5 else line[4:])
+                trailing_blanks = trailing_blanks + 1 if line == '\n' else 0
                 continue
             if not line.replace('\t', '    ', 1).startswith('    '):
                 lines.backstep()
                 break
             line_buffer.append(cls.strip(line))
+            trailing_blanks = 0
+        for _ in range(trailing_blanks):
+            line_buffer.pop()
+            lines.backstep()
         return line_buffer
 
     @staticmethod
@@ -1041,6 +1047,7 @@ class HTMLBlock(BlockToken):
                     break
             elif line.strip() == '':
                 line_buffer.pop()
+                lines.backstep()
                 break
         return line_buffer
 
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -128,7 +128,11 @@ class Document(BlockToken):
     """
     Document token.
     This is a container block token. Its children are block tokens - container or leaf ones.
+
+    Attributes:
+        footnotes (dictionary): link reference definitions.
     """
+
     def __init__(self, lines):
         if isinstance(lines, str):
             lines = lines.splitlines(keepends=True)
@@ -146,15 +150,14 @@ class Heading(BlockToken):
 
     Attributes:
         level (int): heading level.
-        children (list): inner tokens.
     """
     repr_attributes = ("level",)
-    pattern = re.compile(r' {0,3}(#{1,6})(?:\n|\s+?(.*?)(?:\n|\s+?#+\s*?$))')
+    pattern = re.compile(r' {0,3}(#{1,6})(?:\n|\s+?(.*?)(\n|\s+?#+\s*?$))')
     level = 0
     content = ''
 
     def __init__(self, match):
-        self.level, content = match
+        self.level, content, self.closing_sequence = match
         super().__init__(content, span_token.tokenize_inner)
 
     @classmethod
@@ -166,12 +169,13 @@ class Heading(BlockToken):
         cls.content = (match_obj.group(2) or '').strip()
         if set(cls.content) == {'#'}:
             cls.content = ''
+        cls.closing_sequence = (match_obj.group(3) or '').strip()
         return True
 
     @classmethod
     def read(cls, lines):
         next(lines)
-        return cls.level, cls.content
+        return cls.level, cls.content, cls.closing_sequence
 
 class SetextHeading(BlockToken):
     """
@@ -184,8 +188,10 @@ class SetextHeading(BlockToken):
         level (int): heading level.
     """
     repr_attributes = ("level",)
+
     def __init__(self, lines):
-        self.level = 1 if lines.pop().lstrip().startswith('=') else 2
+        self.underline = lines.pop().rstrip()
+        self.level = 1 if self.underline.endswith('=') else 2
         content = '\n'.join([line.strip() for line in lines])
         super().__init__(content, span_token.tokenize_inner)
 
@@ -351,7 +357,6 @@ class BlockCode(BlockToken):
     This is a leaf block token with a single child of type span_token.RawText.
 
     Attributes:
-        children (list): contains a single span_token.RawText token.
         language (str): always the empty string.
     """
     repr_attributes = ("language",)
@@ -402,15 +407,18 @@ class CodeFence(BlockToken):
     This is a leaf block token with a single child of type span_token.RawText.
 
     Attributes:
-        children (list): contains a single span_token.RawText token.
         language (str): language of code block (default to empty).
     """
     repr_attributes = ("language",)
-    pattern = re.compile(r'( {0,3})(`{3,}|~{3,}) *(\S*)')
+    pattern = re.compile(r'( {0,3})(`{3,}|~{3,})( *(\S*)[^\n]*)')
     _open_info = None
+
     def __init__(self, match):
         lines, open_info = match
-        self.language = span_token.EscapeSequence.strip(open_info[2])
+        self.indentation = open_info[0]
+        self.delimiter = open_info[1]
+        self.info_string = open_info[2]
+        self.language = span_token.EscapeSequence.strip(open_info[3])
         self.children = (span_token.RawText(''.join(lines)),)
 
     @property
@@ -423,12 +431,12 @@ class CodeFence(BlockToken):
         match_obj = cls.pattern.match(line)
         if not match_obj:
             return False
-        prepend, leader, lang = match_obj.groups()
+        prepend, leader, info_string, lang = match_obj.groups()
         # info strings for backtick code blocks may not contain backticks,
         # but info strings for tilde code blocks may contain both tildes and backticks.
-        if leader[0] == '`' and '`' in line[match_obj.end(2):]:
+        if leader[0] == '`' and '`' in info_string:
             return False
-        cls._open_info = len(prepend), leader, lang
+        cls._open_info = len(prepend), leader, info_string, lang
         return True
 
     @classmethod
@@ -451,10 +459,9 @@ class CodeFence(BlockToken):
 class List(BlockToken):
     """
     List token.
-    This is a container block token. Its children are list item tokens.
+    This is a container block token. Its children are ListItem tokens.
 
     Attributes:
-        children (list): a list of ListItem tokens.
         loose (bool): whether the list is loose.
         start (NoneType or int): None if unordered, starting number if ordered.
     """
@@ -509,6 +516,12 @@ class ListItem(BlockToken):
     This is a container block token. Its children are block tokens - container or leaf ones.
 
     Not included in the parsing process, but called by List.
+
+    Attributes:
+        leader (string): a bullet list marker or an ordered list marker.
+        prepend (int): the start position of the content, i.e., the indentation required
+                       for continuation lines.
+        loose (bool): whether the list is loose.
     """
     repr_attributes = ("leader", "prepend", "loose")
     pattern = re.compile(r' {0,3}(\d{0,9}[.)]|[+\-*])($|\s+)')
@@ -646,12 +659,11 @@ class ListItem(BlockToken):
 class Table(BlockToken):
     """
     Table token.
-    This is a container block token. Its children are table row tokens.
+    This is a container block token. Its children are TableRow tokens.
 
     Attributes:
-        has_header (bool): whether table has header row.
+        header: header row (TableRow).
         column_align (list): align options for each column (default to [None]).
-        children (list): inner tokens (TableRows).
     """
     repr_attributes = ("column_align",)
     def __init__(self, lines):
@@ -661,6 +673,7 @@ class Table(BlockToken):
             self.header = TableRow(lines[0], self.column_align)
             self.children = [TableRow(line, self.column_align) for line in lines[2:]]
         else:
+            # note: not reachable, because read() guarantees the presence of three dashes
             self.column_align = [None]
             self.children = [TableRow(line) for line in lines]
 
@@ -708,9 +721,12 @@ class Table(BlockToken):
 class TableRow(BlockToken):
     """
     Table row token. Supports escaped pipes in table cells (for primary use within code spans).
-    This is a container block token. Its children are table cell tokens.
+    This is a container block token. Its children are TableCell tokens.
 
     Should only be called by Table.__init__().
+
+    Attributes:
+        row_align (list): align options for each column (default to [None]).
     """
     repr_attributes = ("row_align",)
     # Note: Python regex requires fixed-length look-behind,
@@ -734,7 +750,6 @@ class TableCell(BlockToken):
 
     Attributes:
         align (bool): align option for current cell (default to None).
-        children (list): inner (span-)tokens.
     """
     repr_attributes = ("align",)
     def __init__(self, content, align=None):
@@ -744,11 +759,12 @@ class TableCell(BlockToken):
 
 class Footnote(BlockToken):
     """
-    Footnote token. A "link reference definition" according to the spec.
-    This is a leaf block token. Its children are inline (span) tokens.
+    Footnote token. One or more "link reference definitions" according to the CommonMark spec.
+    This is a leaf block token without children.
 
-    The constructor returns None, because the footnote information
-    is stored in Footnote.read.
+    The constructor returns None, because available footnote definitions are parsed
+    and stored into the root node within `Footnote.read()`. We don't put instances of
+    this class into the resulting AST.
     """
     def __new__(cls, _):
         return None
@@ -800,6 +816,7 @@ class Footnote(BlockToken):
         if not match_info:
             return None
         _, dest_end, dest = match_info
+        dest_type = "angle_uri" if string[dest_start] == "<" else "uri"
 
         # either of:
         # 1) optional spaces or tabs and then a line break to finish the link reference definition.
@@ -817,7 +834,7 @@ class Footnote(BlockToken):
             # we still have a valid link reference definition. otherwise not.
             eol_pos = string[dest_end:title_start].find("\n")
             if eol_pos >= 0:
-                return dest_end + eol_pos + 1, (label, dest, "")
+                return dest_end + eol_pos + 1, (label, dest, "", dest_type, None)
             else:
                 return None
         _, title_end, title = match_info
@@ -826,7 +843,8 @@ class Footnote(BlockToken):
         line_end = title_end
         while line_end < len(string):
             if string[line_end] == '\n':
-                return line_end + 1, (label, dest, title)
+                title_delimiter = string[title_start] if title_start < title_end else None
+                return line_end + 1, (label, dest, title, dest_type, title_delimiter)
             elif string[line_end] in whitespace:
                 line_end += 1
             else:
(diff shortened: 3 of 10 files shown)


test/test_contrib/test_markdown_renderer.py
```diff
@@ -0,0 +1,147 @@
...
+            rendered = renderer.render(Document(input))
+            return rendered
+
+    def test_empty_document(self):
+        input = []
+        rendered = self.roundtrip(input)
+        self.assertEqual(rendered, "".join(input))
+
+    def test_paragraphs_and_blank_lines(self):
+        input = ['Paragraph 1. Single line. Followed by two white-space-only lines.\n',
```
pbodnar, line 20: Tip: for these multi-line inputs, it is probably easier and good for readability to use Python's multi-line strings, together with `dedent()` if we want to keep test code indented - see https://github.com/miyuchina/mistletoe/blob/master/test/test_traverse.py#L28 as an example.

pbodnar:
@anderskaplan, thank you for another contribution. :)

I will try to check it out soon, but of course, I cannot promise anything. Hopefully we'll also get feedback from some other contributors this time. (I think it would be ideal to firstly finish all the current smaller changes (when compared to this one) and finally release version 1.0.0 and this one could go for example into 1.1.0, but we'll see...)

> 7\. The markdown renderer has a main method. It can be convenient, since the renderers in contrib aren't easily
>         run from the regular mistletoe cli. Not sure if it should be kept. (Or whether the markdown renderer should
>         stay in the contrib directory or be part of core?)

Yeah, I'm not a big fan of the "contrib" folder as well, I would probably vote for getting rid of it ASAP, see #101 for more details & discussion.

test/test_contrib/test_markdown_renderer.py
anderskaplan: Yeah, I started out like that, but switched to this format because that's how most of the other test cases in the project were written. Turns out that this format was more readable (at least in my eyes). Especially for test cases like the code blocks where indentation is very irregular -- and important.

pbodnar: Yeah, it seems like a matter of taste... I pretty much like the multi-line format, because it is less typing and generally faster for reading (at least for me ;)). But of course, at minimum for the sake of consistency with most of the existing tests, your choice is just OK as well. :)

karlicoss:
hey, any updates on this? :)

anderskaplan:
Hi @karlicoss, this branch is dependent on #172. As soon as that PR has been merged, I'll be able to prepare this one for showtime.

geolta:
Hi, I had no deep look into the code but I hope that someone else knows the answer: Is implementing a (live) syntax highlighter based on mistletoe possible after these changes? As far as I have seen, it is not the positions that are saved (which would be best for my specific case), but the syntax elements like `**`. But that shouldn't be a problem anyway, since the Markdown renderer should be able to be modified so that the positions can be pulled out of there again, right?

And: Really thank you all for this great Markdown parser!

anderskaplan:
Hi @geolta , it depends a lot what you mean by "(live) syntax highlighter". I suppose "live" means that it highlights while the user is typing. But what is it you want to highlight -- is it for example headings, emphasized text, and the like? And is this for a user interface written in python? If the answer is yes in both cases, then I think mistletoe should be a good match.

pbodnar (changes requested):
@anderskaplan, thanks a lot. I have reviewed the first 2 commits for now, giving you some feedback until I find the time for another review session...

mistletoe/markdown_renderer.py
```diff
@@ -0,0 +1,542 @@
...
+from mistletoe.base_renderer import BaseRenderer
+
+
+class BlankLine(block_token.BlockToken):
+    """
+    Blank line token. Represents a single blank line.
+    This is a leaf block token without children.
+    """
+
+    pattern = re.compile(r"\s*\n$")
```
pbodnar, line 19: Just a detail, but can we please have `'` instead of `"`, generally? `'` seems to be the prevailing variant, and I would like to make the code base consistent in this regard in the near future.

mistletoe/markdown_renderer.py
```diff
@@ -0,0 +1,542 @@
...
+        document is retained as much as possible.
+        """
+        if isinstance(token, block_token.BlockToken):
+            lines = self.render_map[token.__class__.__name__](
+                token, max_line_length=max_line_length
+            )
+        else:
+            lines = self.span_to_lines([token], max_line_length=max_line_length)
+
+        return "".join(map(lambda line: line + "\n", lines))
```
pbodnar, line 130: OK, can you just explain a little bit why we add `\n` explicitly here, while in the `BaseRenderer`, there is just `''.join(map(self.render, token.children))`?

pbodnar (changes requested):
I haven't completed the review yet, getting close slowly :P, here are another few remarks...

test/test_block_token.py
```diff
@@ -125,6 +125,16 @@ def test_parse_indented_code(self):
         arg = 'rm dir\nmkdir test\n'
         self._test_match(block_token.BlockCode, lines, arg, language='')
 
+    def test_parse_indented_code_with_embedded_blank_line(self):
```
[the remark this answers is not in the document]
anderskaplan: Sure - the purpose is to test the behavior when the code block has embedded blank lines, with or without indentation. I agree that one test case is enough, so I'll remove one of them.

mistletoe/block_token.py
```diff
@@ -128,7 +128,12 @@ class Document(BlockToken):
     """
     Document token.
     This is a container block token. Its children are block tokens - container or leaf ones.
+
+    Attributes:
+        children (list): inner tokens.
```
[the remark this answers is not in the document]
anderskaplan: Sure, I can do that. This has gone back and forth a few times, as at first I thought I'd add _all_ the attributes, but then decided not to include the "unofficial" ones. So that sometimes left only the inherited attributes.

mistletoe/block_token.py
```diff
@@ -182,10 +188,14 @@ class SetextHeading(BlockToken):
 
     Attributes:
         level (int): heading level.
+        children (list): inner tokens.
     """
     repr_attributes = ("level",)
+
     def __init__(self, lines):
-        self.level = 1 if lines.pop().lstrip().startswith('=') else 2
+        underline = lines.pop().strip()
```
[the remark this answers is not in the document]
anderskaplan: Sure, saving the whole line instead of `underline_length` does make things a little simpler. The `level` attribute is still needed by HtmlRenderer and others, though.

mistletoe/block_token.py
```diff
@@ -744,8 +776,8 @@ def __init__(self, content, align=None):
 
 class Footnote(BlockToken):
     """
-    Footnote token. A "link reference definition" according to the spec.
-    This is a leaf block token. Its children are inline (span) tokens.
+    Footnote token. A sequence of "link reference definitions" according to the spec.
```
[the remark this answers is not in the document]
anderskaplan: This is tricky. What I do know is that:
- A single Footnote.read() call will parse as many link reference definitions as it can before stopping.
- The Footnote itself won't store any information; it goes into _root_token.footnotes.
- The swap-in replacement I made for Footnote is a LinkReferenceDefinitionBlock (a block token), which in turn contains LinkReferenceDefinitions (a span token) as children. This was the design option that matched the rest of the codebase most closely, and makes both parsing and rendering most straightforward.
- The second removed line was misleading because Footnote tokens do not have children.

The reason for modifying the comment is mainly the first point here: that it will consume all link reference definitions as it can when parsing. All in all I think the updated description is more accurate, so I'd suggest keeping it. Or maybe rewriting it again to something better. Any suggestions?

mistletoe/block_token.py
```diff
@@ -927,7 +961,7 @@ def match_link_title(cls, string, offset):
 
     @staticmethod
     def append_footnotes(matches, root):
-        for key, dest, title in matches:
+        for key, dest, title, _, _ in matches:
```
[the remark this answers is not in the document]
anderskaplan: nice!

mistletoe/span_token.py
```diff
@@ -215,18 +230,21 @@ def strip(cls, string):
...
     repr_attributes = ("soft",)
     pattern = re.compile(r'( *|\\)\n')
     parse_inner = False
     parse_group = 0
 
     def __init__(self, match):
-        content = match.group(1)
-        self.soft = not content.startswith(('  ', '\\'))
-        self.content = ''
+        self.content = ""
```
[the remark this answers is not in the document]
anderskaplan: Yes, I think it might be there for consistency. Another option could be to use `content` instead of introducing a new attribute `marker` -- what do you think?

mistletoe/block_token.py
```diff
@@ -1058,7 +1064,7 @@ def read(cls, lines):
                 if cls._end_cond in line.casefold():
                     break
             elif line.strip() == '':
-                line_buffer.pop()
+                lines.backstep()
```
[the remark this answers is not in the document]
anderskaplan: It doesn't really matter for the HtmlBlock, because the trailing endlines are stripped off in `__init__` anyway. But I'll add it back for consistency.

README.md
```diff
@@ -108,6 +110,18 @@ with open('foo.md', 'r') as fin:
     rendered = mistletoe.markdown(fin, LaTeXRenderer)
 ```
 
+To reflow the text in a Markdown document with a max line length of 20 characters:
+
+```python
+import mistletoe
+from mistletoe.markdown_renderer import MarkdownRenderer
```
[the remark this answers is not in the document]
anderskaplan: Yeah, I think it's nice that you don't _have to_ import the markdown renderer (nor the latex renderer) if you don't want to. I don't mind an explicit import like this.

Maybe the module should be called `markdown` instead of `markdown_renderer`, for shorter imports? But that would break the symmetry with the other renderers. And renaming those modules would be a major compatibility break.

The one place where it would be nicer with a shorter reference path is in the CLI. I mean, a command like this isn't exactly easy to type:
```
mistletoe foo.md --renderer mistletoe.contrib.jira_renderer.JIRARenderer
```

But that could be solved in a different way; for example by letting the CLI know about at least the renderers in the mistletoe main package so that those could be referenced by name without the full path.

mistletoe/markdown_renderer.py
anderskaplan: I have absolutely no preferences about code formatting issues like this, I just like a consistent code base. That's why I'm fond of the [black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#strings) formatter, and the preference there happens to be double quotes.

So, I can change to single quotes, but before I do that: are you sure you want to standardize on single quotes and not double quotes? From what I can see, black is the only popular formatter that standardizes the quotes, and it chose the other alternative.

mistletoe/markdown_renderer.py
anderskaplan: Sure, I'll try!

Here in MarkdownRenderer.render() we can pass in either a block token or a span token for rendering. If it's a block token, we'll look it up in the render_map and find the appropriate rendering method to call. All rendering methods for block tokens return a sequence of lines (without trailing endlines). If, on the other hand, it's a span token, then we call the self.span_to_lines method to render it into a sequence of lines (also without trailing endlines). The string join operation smacks an endline on each line before the join.

Why then don't the rendering methods already include the endlines? That's really just a matter of taste. I think it's cleaner to work with lines without the endlines and apply the endlines in a single place, that's all.

In the BaseRenderer, *all* the token rendering methods (block and span) return plain strings to be joined. Hence the render method can be simpler.

One could argue that this design where the overridden methods in a child class returns a different type of data from the base class isn't exactly a schoolbook example of inheritance. But I think the pros are bigger than the cons in this case and that it's actually pretty neat. The block tokens behave very differently from the span tokens from the perspective of the Markdown renderer. These rendering methods aren't supposed to be called by anyone outside the class. And python, being an untyped language, has no problem with it.

mistletoe/markdown_renderer.py
pbodnar: Well, you forced me to do some extra research again, but in the good way of course. :)

TBH, I'm not still quite sure about this decision. The dilemma is nicely described for example in <https://medium.com/pythoniq/so-single-or-double-quotes-in-python-47c2e7425f32>. Anyway, as I primarily "live" in he Java world, I think I have no problem with using **double quotes** everywhere. Also, as you mentioned, it is the _default_ choice in the popular Black formatter (i.e. it should be a switchable option when needed).

README.md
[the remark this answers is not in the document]
pbodnar: Yeah, lots of topics for future improvements. :) The CLI and its options could definitely be improved. Naming probably as well (I don't like we don't historically follow strict camel case, i.e. `HtmlRenderer` for example). I have also thought about how to pass different renderers' options from the command line. Something that deserves a separate issue(s). That being said, I would leave it as-is for now, agreed? :)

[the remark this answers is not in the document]
anderskaplan: Yes, sounds good. Those are definitively separate issue(s) as you say.

mistletoe/markdown_renderer.py
anderskaplan: With _black_ you only have the option to disable string formatting; the preferred string separator is double quotes and that's it. But ... it turns out that enough people have seen the need for the option, so there is now also a black derivative called _blue_ where the default is single quotes. So I'll reformat with blue, and that should do the trick.

anderskaplan: Maybe I jumped to conclusions too fast there. But it's easy to back out that commit again, should you decide that double quotes is the way to go.

pbodnar: Yeah, I think I'm OK with _black_'s double quotes for now. :)

mistletoe/markdown_renderer.py
pbodnar: OK, thanks for the explanation. So I think we can leave as-is, at least for now, if it doesn't seem to break or "break" anything now or in the near future. :)

pbodnar (changes requested):
Another round complete. :P :)

mistletoe/markdown_renderer.py
anderskaplan: ok! I've backed out the reformatting commit.

mistletoe/block_token.py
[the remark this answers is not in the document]
anderskaplan: I've updated this, but I'm not 100% it ended up the way you intended, so please take another look.

mistletoe/span_token.py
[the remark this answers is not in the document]
anderskaplan: Updated. This also affects the repr() output, but I don't think it's in a bad way.

mistletoe/markdown_renderer.py
pbodnar: OK, thank you.

dev-guide.md
```diff
@@ -244,3 +244,51 @@ with open('foo.md', 'r') as fin:
 For more info, take a look at the `base_renderer` module in mistletoe.
 The docstrings might give you a more granular idea of customizing mistletoe
 to your needs.
+
+## Markdown to Markdown
+
+Suppose you have some Markdown that you want to process and then output
+as Markdown again. Thanks to the text-like nature of Markdown, is is often
```
pbodnar, line 251: typo:

```suggestion
as Markdown again. Thanks to the text-like nature of Markdown, it is often
```

mistletoe/markdown_renderer.py
```diff
@@ -0,0 +1,537 @@
...
+        yield Fragment(token.content, wordwrap=True)
+
+    def render_strong(self, token: span_token.Strong) -> Iterable[Fragment]:
+        return self.embed_span(Fragment(token.delimiter * 2), token.children)
+
+    def render_emphasis(self, token: span_token.Emphasis) -> Iterable[Fragment]:
+        return self.embed_span(Fragment(token.delimiter), token.children)
+
+    def render_inline_code(self, token: span_token.InlineCode) -> Iterable[Fragment]:
+        yield Fragment(token.delimiter + token.raw_content + token.delimiter)
```
pbodnar, line 145: Using `InlineCode.raw_content` here, while it makes sense, is a bit problematic.  Because we need to think about this when altering the AST like in the nice example you prepared in the `dev-guide.md` - i.e. changing inner `RawText` instance has no effect in the example when used with the `MarkdownRenderer`, so results may be surprising.

I think the main complication here is that we decided, I believe correctly, to do some preprocessing of the inline code according to the spec globally, namely this line within `InlineCode.__init__()`:

```python
content = content.replace('\n', ' ')
```

Any ideas of how to get out of this "conflicting stuff" elegantly?

mistletoe/markdown_renderer.py
```diff
@@ -0,0 +1,537 @@
...
+    def span_to_lines(
+        self, tokens: Iterable[span_token.SpanToken], max_line_length: int = None
+    ) -> Iterable[str]:
+        """
+        Renders a sequence of span (inline) tokens into a sequence of lines.
+        """
+        fragments = self.make_fragments(tokens)
+        return self.fragments_to_lines(fragments, max_line_length=max_line_length)
+
+    def make_fragments(self, tokens: Iterable[span_token.SpanToken]):
```
pbodnar, line 382: I.e. we could add this?

```suggestion
    def make_fragments(self, tokens: Iterable[span_token.SpanToken]
    ) -> Iterable[Fragment]:
```

mistletoe/markdown_renderer.py
```diff
@@ -0,0 +1,537 @@
...
+        If `max_line_length` is specified, the document is word wrapped to the
+        specified line length. Otherwise the formatting from the original (parsed)
+        document is retained as much as possible.
+        """
+        if isinstance(token, block_token.BlockToken):
+            lines = self.render_map[token.__class__.__name__](
+                token, max_line_length=max_line_length
+            )
+        else:
+            lines = self.span_to_lines([token], max_line_length=max_line_length)
```
pbodnar, line 128: While it seems it to work, I need to think about some simplifications here. What bothers me is that we have to know / pass _everywhere_ in the nested (render) functions:

1. Whether children are block, or span tokens - i.e. to distinguish calling different (helper) functions. Quick idea: Cannot we generalize this somehow? For example to define a "good old" `render_inner()` which would handle the if-then-else in a similar fashion as can be seen here? Or to have Fragments returned even from blocks rendering functions?

2. The options - now it is just `max_line_length`, but what if there will be more? Quick idea: What about storing the options in `self` (passing it to the constructor already?)?

pbodnar (changes requested):

pbodnar:
It's been a while since the beginning of this PR. I think that we could just finish the minor cleanups and put this long awaited feature into the very next release version, with a warning that the renderer's API is in a "preview" state and it might change in a future version. What do you think?

anderskaplan:
Sorry for the slow responses lately @pbodnar -- It's been a very busy May. 

I fully agree that it would be very nice to get this into the next major release, and I'll try and resolve the remaining comments as soon as possible.

mistletoe/markdown_renderer.py
```diff
@@ -0,0 +1,537 @@
...
+            *chain(
+                (
+                    block_token.HTMLBlock,
+                    span_token.HTMLSpan,
+                    BlankLine,
+                    LinkReferenceDefinitionBlock,
+                ),
+                extras,
+            )
+        )
```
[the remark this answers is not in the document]
anderskaplan: It's not only readability: the way the parsers are initialized and the fact that you can't have two different parsers initialized at the same time (because they will overwrite each others' choice of block and span tokens) is a problem. But it will take a major API break to fix it -- unless there's some clever solution that I haven't thought of.

mistletoe/markdown_renderer.py
anderskaplan: For 1, I think this is less of a problem than what it might seem like. The `render` method is the _only_ method in the class where the check is needed, because in all the rendering methods we already know whether the children are block or span tokens. A token will only have children of one or the other kind.

The `render_inner`-like method you're looking for is probably the dynamic duo consisting of `blocks_to_lines` and `span_to_lines`.

anderskaplan: For 2, sure, it could become an issue. And I see that some of the other renderers have options in the constructors, so that seems like a reasonable approach.

anderskaplan:
This question should also be addressed before merging, imho:

> Four new tokens! An open question is whether they belong in the same module as the markdown renderer class, or if they should be moved somewhere else. Maybe create a new module to host all optional tokens?

I think moving all optional tokens into a module could be pretty neat. What do you say @pbodnar ?

mistletoe/markdown_renderer.py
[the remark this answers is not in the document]
pbodnar: Yeah, sure, that's the problem with using the various global variables like `_token_types` in mistletoe you are talking about, right? The "syntactic sugar refactoring" suggested above won't solve that on its own for sure.

Related: I was thinking about those renderers' constructor parameters - what about introducing explicitly named parameter `extra_tokens` instead of that current `*extras`? That would let us pass dynamically constructed options (e.g. read from the command line) to the constructors. But of course, it is an API change again. But I think we won't avoid that anyway, sooner or later...

mistletoe/markdown_renderer.py
pbodnar: Ad 1 - I agree, technically, it is probably not that big problem, but I'm still unsure if it won't complicate some stuff later on. I would say "will see" for now.

Ad 2 - Yes, storing the options in the constructor is the only feasible way I know of currently. I was also thinking about some globally set-able "mistletoe.options", but making something global seems risky and it would also be behind the scope of this PR I guess. ;)

pbodnar:
> This question should also be addressed before merging, imho:
> 
> > Four new tokens! An open question is whether they belong in the same module as the markdown renderer class, or if they should be moved somewhere else. Maybe create a new module to host all optional tokens?
> 
> I think moving all optional tokens into a module could be pretty neat. What do you say @pbodnar ?

Or maybe 3 new tokens currently if I look correctly (`Fragment` class not included)? I think I like them where they are now, but maybe if you give me a better idea of for example their reuse from other (future) modules, I could change my mind? :)

mistletoe/markdown_renderer.py
anderskaplan: I agree that it's a problem -- it would be much better to use the content of the RawText node instead of this new and rather hidden field. Only, we can't do that if the RawText content is processed.

The preprocessing in `InlineCode.__init__()` consists of two steps: first we convert line breaks to space, and then we strip one leading and one trailing space if the string begins AND ends with spaces and is not a spaces-only string. This makes the Markdown to HTML rendering spec-compliant, so of course the processing is needed in that process.

One option is to move the processing into `HTMLRenderer.render_inline_code()`. The other renderers (LaTeX, Jira, XWiki20, AST, Scheme) will also need to be updated accordingly -- I don't know what the expected behavior is for them, but that should be possible to find out (and it's easy to just keep the current behavior). It will be an API break but not a severe one.

Elegant or not, this is the least bad solution I can think of :)

mistletoe/markdown_renderer.py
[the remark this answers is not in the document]
anderskaplan: Yes, that's exactly it. I think renaming "extras" to "extra_tokens" and opening up for a way to specify options from the CLI sounds like a good idea.

mistletoe/markdown_renderer.py
anderskaplan: I'll go with the options in the constructors for now.

mistletoe/markdown_renderer.py
pbodnar: Good points, thank you. Or, I have got an idea now that we could temporarily disable the preprocessing of inline code - within the `with MarkdownRenderer() as renderer` block (by overriding the `__enter__` and `__exit__` methods, plus introducing a special boolean flag variable). That could do the trick? :)

anderskaplan: Solved the issue by allowing inline code to be re-flowed, that is, not preserving the line breaks. While this is against the principle of preserving the formatting as far as possible, it is more in line with how span tokens work in general, and with how inline code is handled when rendering to HTML.

An option could be to allow the InlineToken to hold a sequence of RawText and LineBreak tokens, instead of a single RawText node. (Possible future improvement?)

anderskaplan:
There, I believe all comments are addressed now. 😄

mistletoe/markdown_renderer.py
```diff
@@ -0,0 +1,544 @@
...
+        self, token: block_token.Document, max_line_length: int
+    ) -> Iterable[str]:
+        return self.blocks_to_lines(token.children, max_line_length=max_line_length)
+
+    def render_heading(
+        self, token: block_token.Heading, max_line_length: int
+    ) -> Iterable[str]:
+        # note: no word wrapping, because atx headings always fit on a single line.
+        line = "#" * token.level
+        text = next(self.span_to_lines(token.children, max_line_length=None), "")
```
pbodnar, line 254: OK, even if I still think we could omit the passing of `max_line_length` all the way down when we've got it stored in `self` now (and specifically handle just the extra cases where wrapping is not allowed - like in here). But OK, good enough for now, we can refactor later on if there will be time. ;)

pbodnar (changes requested):
@anderskaplan, thanks a lot. Just 1 or 2 smaller remarks which I think we could look at before merging.

mistletoe/markdown_renderer.py
anderskaplan: The reason for the parameter to still be there in the method calls is that the effective line length changes during rendering. For example, if you're rendering a block quote, then each line is prefixed with a `> `, and the effective max line length decreases by two. So while the naming can be discussed, it's not really possible to use the value stored with `self`.

pbodnar: Oh, I see, makes sense. This reminds me of I haven't done a full review of all the code yet. :') I'm afraid I don't have enough capacity for that many lines now, so I guess I will just risk it and merge it without a detailed review.

pbodnar (changes requested):
@anderskaplan, thanks for another round. :)

Do you think you can prepare a final set of squashed atomic commits, ideally with commit messages as described at https://github.com/miyuchina/mistletoe/blob/master/CONTRIBUTING.md? At minimum, I would like to have all summary lines end with ` (#162)`, which I believe is pretty practical think. Sure, I can do this myself, but this is still kinda of your "child", so... :)

anderskaplan:
Here we go!

pbodnar (approved):
@anderskaplan, good job, thank you once again. :) If you don't mind, I will give you explicit credits in the release notes.

I have also run `python test/benchmark.py mistletoe` to see performance impact of this PR. It looks like there is about 1 to 2 % performance penalty due to the extended parsing. Nothing to worry about then.

PR #167, opened 2022-12-03
Contrib folder update

anderskaplan:
This PR is a proposed solution for #101, Make renderers from the "contrib" folder easier to use.
The approach taken is as described in the discussion thread on that issue: basically to move the contrib directory inside the mistletoe source directory.

In addition to moving files around, the PR also includes some documentation updates to explain the use of the contrib folder and how to use the renderers in the contrib directory from the CLI.

NOTE the mistletoe.egg-info folder has *not* been updated, as the egg format seems to be [quite obsolete by now](https://stackoverflow.com/questions/2051192/what-is-a-python-egg) and superseeded by [wheel](https://packaging.python.org/en/latest/discussions/wheel-vs-egg/). *I'd recommend to remove the egg-info folder.*


tucked (approved):
Woohoo 🎉 LGTM! Thanks for this 👍

pbodnar:
> NOTE the mistletoe.egg-info folder has _not_ been updated, as the egg format seems to be [quite obsolete by now](https://stackoverflow.com/questions/2051192/what-is-a-python-egg) and superseeded by [wheel](https://packaging.python.org/en/latest/discussions/wheel-vs-egg/). _I'd recommend to remove the egg-info folder._

Well, the ".egg-info" folder is _not_ versioned and it seems to be generated when installing mistletoe via `pip3 install -e .`. So no problem here, actually? :)

pbodnar:
> ... Added __init__.py to make the contrib directory a python module. ...

I guess this is OK. I have found that the `__init__.py` file is no longer necessary since Python 3.3, yet it seems to be recommended for "regular packages", like this one - as described at [SO](https://stackoverflow.com/questions/37139786/is-init-py-not-required-for-packages-in-python-3-3/48804718#48804718)?

pbodnar (changes requested):
@anderskaplan, thank you! I think it looks fine, just some smaller amendments left. :)

Also, I expect that I can probably squash all the commits into one on merge, while adding issue link (#101) to the commit message, OK?

pbodnar:
@anderskaplan, LGTM, thank you!

I just have to prepare a PR for changing https://github.com/hluk/copyq-commands/blob/master/Global/convert-markdown.ini which uses the renderers from the contrib folder, and I will merge this afterwards.

Also, what do you think about including this "breaking change" (not much we can do to keep it backwards compatible at least for a while, I guess) into version 1.0.0 - yes, or no? :)

anderskaplan:
Hi @pbodnar, as you say there's not much that can be done to keep this backward compatible, so I thinks its better to rip off the band-aid as soon as possible and be done with it. Besides, the label "contrib" indicates that the software is more experimental and that a lower standard of backward compatibility is to be expected.

anderskaplan:
Rebased to resolve a merge conflict.

pbodnar (approved):

pbodnar:
Hi @anderskaplan, thank you, so I have merged it right away. :)

Issue #163, opened 2022-09-21
Question: Inconsistencies in the block tokens


anderskaplan:
There are some inconsistencies among the block tokens that maybe should be fixed before stepping up to version 1.0:

1. Trailing newlines are sometimes preserved and sometimes not. `CodeFence` and `BlockCode` preserve them; `Paragraph` and `HTMLBlock` do not.
2. `CodeFence` and `BlockCode` keep their content in a single `RawText` child node, whereas the `HTMLBlock` keeps it in the `content` property. In fact, the `HTMLBlock` is the only block token to have a `content` property. It is typically used with span tokens.

So what to do about it?

My suggestion would be to remove the trailing newlines from all block tokens. The other consistent option, to keep them for all block tokens, would add a trailing `LineBreak` to all `Paragraph`'s, and that would just be a pain. Of course there's also the option to leave it as it is.

I would also suggest to place the `HTMLBlock` content in a single `RawText` node, so it would be consistent with the other block tokens. Maybe keep its `content` property, too, in order to not break the API. The `content` property could be turned into a property getter and marked as deprecated.

Thoughts?

pbodnar:
The suggestions seem reasonable, I will try to review them soon.

I guess there is an implicit expectation that suggested changes would change the output from renderers, making them possibly more consistent, isn't it?

anderskaplan:
The proposal will mainly affect the AST. I think the output from the renderers would only change very little, if at all. But it would be easier to make their output more consistent, too, if desired.

PR #169, opened 2022-12-03
Modified the HTMLBlock token as described in #163, Inconsistencies in the block tokens

anderskaplan:
Modified the HTMLBlock to store its content in a RawText child node, like all the other block tokens, instead of a content attribute.  This makes it easier to iterate over the token tree, because all block tokens can be handled the same way.

Added a convenience wrapper to preserve backward compatibility.

### DIFF
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -361,7 +361,7 @@ class BlockCode(BlockToken):
 
     @property
     def content(self):
-        """Return the code block content."""
+        """Returns the code block content."""
         return self.children[0].content
 
     @staticmethod
@@ -415,7 +415,7 @@ class CodeFence(BlockToken):
 
     @property
     def content(self):
-        """Return the code block content."""
+        """Returns the code block content."""
         return self.children[0].content
 
     @classmethod
@@ -972,7 +972,7 @@ class HTMLBlock(BlockToken):
 
     @property
     def content(self):
-        """Return the raw HTML content."""
+        """Returns the raw HTML content."""
         return self.children[0].content
 
     @classmethod
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -946,10 +946,10 @@ class ThematicBreak(BlockToken):
 class HTMLBlock(BlockToken):
     """
     Block-level HTML token.
-    This is a leaf block token without children.
+    This is a leaf block token with a single child of type span_token.RawText.
 
     Attributes:
-        content (str): the raw HTML content.
+        children (list): contains a single span_token.RawText token with the raw HTML content.
     """
     _end_cond = None
     multiblock = re.compile(r'<(pre|script|style|textarea)[ >\n]')
@@ -958,7 +958,12 @@ class HTMLBlock(BlockToken):
                                 span_token._closing_tag)) + r')\s*$')
 
     def __init__(self, lines):
-        self.content = ''.join(lines).rstrip('\n')
+        self.children = (span_token.RawText(''.join(lines).rstrip('\n')),)
+
+    @property
+    def content(self):
+        """Return the raw HTML content."""
+        return self.children[0].content
 
     @classmethod
     def start(cls, line):
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -359,6 +359,11 @@ class BlockCode(BlockToken):
         self.language = ''
         self.children = (span_token.RawText(''.join(lines).strip('\n')+'\n'),)
 
+    @property
+    def content(self):
+        """Return the code block content."""
+        return self.children[0].content
+
     @staticmethod
     def start(line):
         return line.replace('\t', '    ', 1).startswith('    ')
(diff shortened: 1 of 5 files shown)


pbodnar:
> Also fixed a bug in Token.repr which caused class attributes on the tokens to be included and not only instance attributes.

Good catch, although I can imagine having a separate PR for that. But I know it's tempting to avoid some bureaucracy around creating new branch and PR - I have done that recently myself on another project... ;)

anderskaplan:
Now this PR is really going out on a limb :)

Adding another small change related to the Token.repr issue, which is to unify how Token.repr and the AST renderer select the attributes to be included in their output. I think it makes sense that only "officially declared" attributes should appear in both of those cases, and I see no reason why they should differ in their selection of attributes.

anderskaplan:
Btw, I've looked into the other part of issue/question #163 about inconsistent line endings and come to the conclusion that it's best to leave them as they are. So I consider this PR finished here.

anderskaplan:
The changes not related to the HTMLBlock are now split out into a separate PR. I'll update this PR once those changes are in, because of dependencies.

mistletoe/block_token.py
```diff
@@ -958,7 +958,16 @@ class HTMLBlock(BlockToken):
...
     def __init__(self, lines):
-        self.content = ''.join(lines).rstrip('\n')
+        self.children = (span_token.RawText(''.join(lines).rstrip('\n')),)
+
+    @property
+    def content(self):
+        """Return the raw HTML content.
+
+        Convenience wrapper to preserve backward compatibility.
+        New code should use the `children` attribute instead.
```
anderskaplan, line 968: Hmm, now that I think of it, the client code looks better when you use the property getter. Maybe we shouldn't discourage its use, but instead add similar property getters on the other leaf block tokens, too?

test/test_html_renderer.py
```diff
@@ -127,9 +128,9 @@ def test_thematic_break(self):
         self._test_token('ThematicBreak', '<hr />', children=False)
 
     def test_html_block(self):
-        content = output = '<h1>hello</h1>\n<p>this is\na paragraph</p>\n'
-        self._test_token('HTMLBlock', output,
-                         children=False, content=content)
+        output = '<h1>hello</h1>\n<p>this is\na paragraph</p>\n'
+        token = namedtuple('MockHtmlBlockToken', 'children')([namedtuple('MockRawTextToken', 'content')(output)])
```
pbodnar, line 132: Clever use of `namedtuple()`. :) Maybe just the `Token` is redundant in the names, but that is a detail. ;)

mistletoe/block_token.py
pbodnar: Well, good question. I think we could safely put the `content` property getter also into `BlockCode` and `CodeFence` as suggested, because besides `HTMLBlock` these are the only leaf block tokens which should always have just one `RawText` child, right? Of course, we should be careful when introducing new API like this - I can't see no gotchas here though, can you?

pbodnar (commented):
@anderskaplan, I have finally found some time for feedback, thanks for pushing it further. :)

mistletoe/block_token.py
anderskaplan: Yes, I think it should be perfectly safe, and make for a better API. I'll go ahead and add those.

test/test_html_renderer.py
anderskaplan: Thanks :) However, I've backed out this change to stick with the same pattern as the other test cases.

mistletoe/block_token.py
```diff
@@ -359,6 +359,11 @@ def __init__(self, lines):
         self.language = ''
         self.children = (span_token.RawText(''.join(lines).strip('\n')+'\n'),)
 
+    @property
+    def content(self):
+        """Return the code block content."""
```
pbodnar, line 364: OK, just a slight detail for consistency:

```suggestion
        """Returns the code block content."""
```

DITTO below.

test/test_block_token.py
```diff
@@ -119,13 +119,47 @@ def test_code_fence_with_backticks_and_tildes_in_the_info_string(self):
         self._test_match(block_token.CodeFence, lines, arg, language='aa')
 
 
+class TestCodeFenceBehavior(unittest.TestCase):
```
pbodnar, line 122: Tests seem OK, yet I would simplify the nesting: below, `test_content_property()` is nested within `TestHTMLBlock`, I would nest all analogous `test_content_property()` like this.

pbodnar (changes requested):

test/test_block_token.py
anderskaplan: I moved all the property getter unit tests to a dedicated class. Is that how you meant?

pbodnar: Not quite like that, but how you did it, I think it is OK. Of course, there is always room for some de-duplication, but I think it is good enough for now. :)

mistletoe/block_token.py
anderskaplan: Ok, fixed

test/test_block_token.py
anderskaplan: Ok! So, with that I think all comments should be resolved.

mistletoe/block_token.py
pbodnar: Done, OK.

pbodnar (approved):

pbodnar:
Squashed and merged. Thank you! :)

PR #170, opened 2022-12-09
Include any custom HTML attributes on any Token

hyperking:
This is similar to this [PR#134](https://github.com/miyuchina/mistletoe/pull/134)

I would like to propose an alternative solution that will not alter existing `render` methods.

**Html attribute block Proposed Spec**
Prepend line containing `{...}` will describe the html attributes that will be added during the tokenizer process.
Contents within the Html attribute block will be a comma separated list of key/value pairs. 
Any key prefixed with `>` will apply to the children token.

**Example Html attribute block**
INPUT
```
{id:my-value, class:some-class}
# Mistletoe is Awesome

{id:my-list, class:foo, >class:bar-items}
- Item One
- Item Two
- Item Three\
```
OUTPUT
```
<h1 id="my-value" class="some-class">Mistletoe is Awesome</h1>
<ul id="my-list" class="foo">
    <li class="bar-items">Item One</li>
    <li class="bar-items">Item Two</li>
    <li class="bar-items">Item Three</li>
</ul>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,36 @@ without touching any of the core components.
 
 Remember to spell mistletoe in lowercase!
 
+In This Fork
+------------
+
+**Html attribute block Proposed Spec**
+Prepend line containing `{...}` will describe the html attributes that will be added during the tokenizer process.
+Contents within the Html attribute block will be a comma separated list of key/value pairs. 
+Any key prefixed with `>` will apply to the children token.
+
+
+**Example Html attribute block**
+INPUT
+```
+{id:my-value, class:some-class}
+# Mistletoe is Awesome
+
+{id:my-list, class:foo, >class:bar-items}
+- Item One
+- Item Two
+- Item Three\
+```
+OUTPUT
+```
+<h1 id="my-value" class="some-class">Mistletoe is Awesome</h1>
+<ul id="my-list" class="foo">
+    <li class="bar-items">Item One</li>
+    <li class="bar-items">Item Two</li>
+    <li class="bar-items">Item Three</li>
+</ul>
+```
+
 Features
 --------
 * **Fast**:
--- a/mistletoe/block_tokenizer.py
+++ b/mistletoe/block_tokenizer.py
@@ -74,6 +74,35 @@ def tokenize_block(iterable, token_types):
         line = lines.peek()
     return parse_buffer
 
+def process_props(result: str):
+    if not isinstance(result, str): return None
+    if not result.strip().endswith('}'): return None
+    def get_props(prop):
+        if ':' in prop:
+            key, _, value = prop.partition(":")
+            return f' {key}="{value}"'
+        return None
+    propstr = result.strip().replace('{','').replace('}','')
+    propslst = propstr.split(',')
+    childattr = None
+    parentattr = ''
+
+    for prop in propslst:
+        prop = prop.strip()
+        if prop.startswith('>'):
+            childattr = get_props(prop.lstrip('>'))
+        else:
+            parentattr += get_props(prop)
+    
+    return (parentattr, childattr)
+
+def apply_props(token, props):
+    if not props: return
+    token.html_props = props[0]
+    if not token.children: return
+    for chld in token.children:
+        chld.html_props = props[1]
+
 
 def make_tokens(parse_buffer):
     """
@@ -84,10 +113,16 @@ def make_tokens(parse_buffer):
     and span-level parsing is started here.
     """
     tokens = []
+    props = None
     for token_type, result in parse_buffer:
+        if not props: 
+            props = process_props(result[0])
+            if props: continue
         token = token_type(result)
         if token is not None:
+            apply_props(token, props)
             tokens.append(token)
+            props = None
     return tokens
 
 
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -99,9 +99,11 @@ class HTMLRenderer(BaseRenderer):
         return token.content
 
     def render_heading(self, token: block_token.Heading) -> str:
-        template = '<h{level}>{inner}</h{level}>'
+        template = '<h{level}{attr}>{inner}</h{level}>'
         inner = self.render_inner(token)
-        return template.format(level=token.level, inner=inner)
+        id = inner.replace(' ','-').lower()
+        attr = f' id="{id}"' if not token.html_props else token.html_props
+        return template.format(level=token.level, attr=html.unescape(attr), inner=inner)
 
     def render_quote(self, token: block_token.Quote) -> str:
         elements = ['<blockquote>']
@@ -132,7 +134,7 @@ class HTMLRenderer(BaseRenderer):
             attr = ' start="{}"'.format(token.start) if token.start != 1 else ''
         else:
             tag = 'ul'
-            attr = ''
+            attr = '' if not token.html_props else token.html_props
         self._suppress_ptag_stack.append(not token.loose)
         inner = '\n'.join([self.render(child) for child in token.children])
         self._suppress_ptag_stack.pop()
@@ -148,7 +150,8 @@ class HTMLRenderer(BaseRenderer):
                 inner_template = inner_template[1:]
             if token.children[-1].__class__.__name__ == 'Paragraph':
                 inner_template = inner_template[:-1]
-        return '<li>{}</li>'.format(inner_template.format(inner))
+        attr = '' if not token.html_props else token.html_props
+        return '<li{attr}>{}</li>'.format(inner_template.format(inner), attr=attr)
 
     def render_table(self, token: block_token.Table) -> str:
         # This is actually gross and I wonder if there's a better way to do it.
--- a/mistletoe/token.py
+++ b/mistletoe/token.py
@@ -39,7 +39,7 @@ class Token:
     """
 
     repr_attributes = ()
-
+    html_props = None
     def __repr__(self):
         output = "<{}.{}".format(
             self.__class__.__module__,


anderskaplan:
Hi, a few comments from someone who's been messing a bit with this code base recently.

I think the approach with a new block token is a good one, and it fits nicely in the markdown syntax.
It definitely has an advantage over the other proposed method to add attributes, in that you don't need to
write a custom renderer to use it.

Here are a few suggestions:
1. The way this is added into block_tokenizer.make_tokens() feels a bit hacky. Why not treat it like
   any other block token and handle it in HTMLRenderer.render_html_attributes() instead?
   That's more in line with the extensible design of mistletoe.
2. Since this is optional and non-standard functionality, I'd suggest to not add it to HTMLrenderer but rather to create a new
   renderer class for it in the contrib folder, derived from HTMLRenderer of course. The new
   block token class could go there as well.
3. The format for the `${...}` strings should be defined more clearly. What about spaces, newlines, escaping, ...?
   Not saying it needs to be more complicated than necessary -- just that it should be well defined.
4. Markdown can get complicated and I'm not sure if the proposed design will work well with lists nested inside lists,
   for example. It would be nice with some test cases to demonstrate how it handles more difficult cases. For example,
   if you set the id attribute on a list that has a nested list inside it, you don't want the same id on the inner list
   because that would be invalid html.

Hope this can be helpful to make it a good addition to mistletoe!

hyperking:
> Hi, a few comments from someone who's been messing a bit with this code base recently.

**Thank you for the response and suggestions.**

> 3. The format for the `${...}` strings should be defined more clearly. What about spaces, newlines, escaping, ...?
>    Not saying it needs to be more complicated than necessary -- just that it should be well defined.

Using the `{...}` characters was causing conflicts with other renderers namely the `test_XWiki20Renderer.py` macros test failing tests. Although your suggestion to create a new HTMLrenderer may help fix those issues here.

Per the block syntax. Lets get inline working first before multi-line. I'd imagine we could repurpose the code block functionality

> 4. Markdown can get complicated and I'm not sure if the proposed design will work well with lists nested inside lists,
>    for example. It would be nice with some test cases to demonstrate how it handles more difficult cases. For example,
>    if you set the id attribute on a list that has a nested list inside it, you don't want the same id on the inner list
>    because that would be invalid html.

Great point, `id` attributes should always be unique to pass accessibility standards. We could add a numeric suffix onto the nest element id each time props are passed down from parents. I could also experiment with making the `apply_props` recursive for child tokens.

hyperking:
I've updated the branch to include adding unique id attribute values for deeply nested lists.

Below is an example currently in place
INPUT

```markdown
${id:todos, tabindex:100 > class:list-item}
- Item One
- Item Two
    - item two.one
    - item two.two
        - apples
        - oranges
- Item Three
```

OUTPUT

```html
<ul id="todos" tabindex="100">
    <li class="list-item" tabindex="1">Item One</li>
    <li class="list-item" tabindex="1">Item Two
        <ul id="todos-2" tabindex="1">
            <li class="list-item" tabindex="1">item two.one</li>
            <li class="list-item" tabindex="1">item two.two
                <ul id="todos-2-3" tabindex="1">
                    <li class="list-item" tabindex="1">apples</li>
                    <li class="list-item" tabindex="1">oranges</li>
                </ul>
            </li>
        </ul>
    </li>
    <li class="list-item" tabindex="1">Item Three</li>
</ul>
```

anderskaplan:
You could also consider to let the attributes apply only to the immediately following block token. Then it would be possible to add another set to the inner list, like so:

```
${id:todos}
- Item One
- Item Two
    - item two.one
    - item two.two
        ${id:shopping-list}
        - apples
        - oranges
- Item Three
```

hyperking:
> You could also consider to let the attributes apply only to the immediately following block token. Then it would be possible to add another set to the inner list, like so:
> 

Hmm, I had that idea in the beginning, however the current tokenization processing was a struggle and we need a system to pass parent props to children.

Or perhaps an Emmet style syntax at the root level may function better.

🐊 `${id:todos > id:shopping-list > class:shopping-list-items}`

hyperking:
@anderskaplan There is still a few things I need to update in this PR and I'd like to know what is the criteria for getting code merged into master? Thanks for your help and feedback.

**things todo:**

1. Implement attributes for all other html elements
2. Add test cases for each html element
3. Update Readme? ( not sure If the readme should include this feature )

anderskaplan:
Ok, that's a question for @pbodnar. I can only help with some guidance on how to make the code fit in with the overall design of the codebase.

pbodnar (changes requested):
> @anderskaplan There is still a few things I need to update in this PR and I'd like to know what is the criteria for getting code merged into master? Thanks for your help and feedback.

@hyperking, thank you for your contribution (and thanks to @anderskaplan for giving some useful feedback :)). I have gone through it quickly and gave you some feedback. There are no exactly described criteria. Generally, the code should be production-ready (or "fully baked" ;)) before being merged into master.

It also helps if commits and their messages are already as close to the final state as possible. So rebasing (squashing etc.) and force-pushing is usually OK to achieve that, even though this may lead to losing links from conversations to the code...

---

Some further quick thoughts (as I don't have much time for maintaining mistletoe these days and some open PRs are still waiting there for me ;)):

1) PR description

> This is similar to this https://github.com/miyuchina/mistletoe/pull/134
> 
> I would like to propose an alternative solution that will not alter existing render methods.

OK, it would be also good to highlight that unlike #134, this PR lets _regular users_ specify the additional attributes, so it is not just an API extension.

2) Selecting the syntax for attributes

> Using the `{...}` characters was causing conflicts with other renderers namely the `test_XWiki20Renderer.py` macros test failing tests. Although your suggestion to create a new HTMLrenderer may help fix those issues here.

Provided we would want to use just `{...}` instead of `${...}` and provided a test is failing, just fixing that test could be a solution?

Anyway, I wonder if we couldn't inspire at some alternative Markdown parsers or elsewhere, regarding the syntax (e.g. [XWiki](https://www.xwiki.org/xwiki/bin/view/Documentation/UserGuide/Features/XWikiSyntax/) uses `(% ... %)`)? Also from many tools and languages, I am quite used to read `${...}` as something that should be _evaluated_, that's why I would like to consider some other variants.

3) `HTMLAttributes`, or `ExtraAttributes`?

Any thoughts on making this mechanism, or token, general? I.e. the attributes maybe don't have to be necessarily just for HTML?

Also, `HTMLAttributes` are being added into `block_token.__all__` by this PR now, so they would be already available to all the other renderers, right?

And a related question: shouldn't we do this feature / parsing optional (thinking about performance and backwards compatibility here)?

test/test_html_attributes_renderer.py
```diff
@@ -0,0 +1,35 @@
+from unittest import TestCase
+
+
+class TestHTMLAttributes(TestCase):
+
+    def test_document(self):
```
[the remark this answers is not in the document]
hyperking: 😅 I knew this moment would come. Sure thing!

hyperking:
> @hyperking, thank you for your contribution (and thanks to @anderskaplan for giving some useful feedback :)). I have gone through it quickly and gave you some feedback. There are no exactly described criteria. Generally, the code should be production-ready (or "fully baked" ;)) before being merged into master.

🤓 Awesome! With the holidays underway I may have some in between time to address any loose ends with this PR.

> It also helps if commits and their messages are already as close to the final state as possible. So rebasing (squashing etc.) and force-pushing is usually OK to achieve that, even though this may lead to losing links from conversations to the code...

🤓 Yes, I'll make an effort to clean up the PR as close to `upstream/master` as possible without any conflicts. I've made a test run locally and git didn't seem to complain.

> 2. Selecting the syntax for attributes
> Provided we would want to use just `{...}` instead of `${...}` and provided a test is failing, just fixing that test could be a solution?

🤓  I agree, the original `{ ... } ` syntax would be ideal. However it does conflict with other aforementioned renders. I did make this renderer configurable so that the end user could specify an alternative character literals.

[See configure statement here](https://github.com/hyperking/mistletoe/blob/656ba0a0f01e0ca2b9a0eb4b6d8bdddbe86f002f/mistletoe/block_token.py#L1029-L1034)


> 3. `HTMLAttributes`, or `ExtraAttributes`?
> Any thoughts on making this mechanism, or token, general? I.e. the attributes maybe don't have to be necessarily just for HTML?

🤓 HTMLAttributes are the common term used in web development for describing attributes on a given DOM node. per this feature, It's currently a general token. The token currently only supports basic markdown that renders HTML dom nodes.

> Also, `HTMLAttributes` are being added into `block_token.__all__` by this PR now, so they would be already available to all the other renderers, right?

🤓 Yes, It's also dependent on if the qualifying token string is present in the markdown ( outside of any code blocks )

> And a related question: shouldn't we do this feature / parsing optional (thinking about performance and backwards compatibility here)?

I'd love to see how this performs, but generally this feature would come at a cost, but Its optional though 🤷‍♂️

mistletoe/base_renderer.py
```diff
@@ -41,9 +41,11 @@ class BaseRenderer(object):
                         parsing process.
     """
     _parse_name = re.compile(r"([A-Z][a-z]+|[A-Z]+(?![a-z]))")
+    auto_header_ids = False
```
hyperking, line 44: Note to self: Place within `HTMLAttributesRenderer` class as a configurable property

mistletoe/html_attributes_renderer.py
```diff
@@ -0,0 +1,256 @@
...
+from itertools import chain
+from urllib.parse import quote
+from mistletoe import block_token
+from mistletoe import span_token
+from mistletoe.block_token import HTMLBlock
+from mistletoe.span_token import HTMLSpan
+from mistletoe.base_renderer import BaseRenderer
+
+
+class HTMLAttributesRenderer(BaseRenderer):
```
anderskaplan, line 16: Consider inheriting from the HTMLRenderer rather than the BaseRenderer. Maybe you can remove some code duplication that way?

mistletoe/html_attributes_renderer.py
```diff
@@ -0,0 +1,256 @@
...
+
+    See mistletoe.base_renderer module for more info.
+    """
+    def __init__(self, *extras):
+        """
+        Args:
+            extras (list): allows subclasses to add even more custom tokens.
+        """
+        self._suppress_ptag_stack = [False]
+        super().__init__(*chain((HTMLBlock, HTMLSpan), extras))
```
anderskaplan, line 28: This is probably where you'll want to insert your HTMLAttributes token class rather than in the BaseRenderer. As you can see, the other HTML tokens aren't referenced from the base renderer either.

On the other hand, if the HTMLAttributes is changed into something more generic then maybe the BaseRender is the right place after all. My preference would be to keep it purely optional i.e. not in the BaseRenderer. That way the performance hit is zero when the functionality isn't used, and will also be no unintended side effects when parsing markdown that happens to have HTMLAttributes-like tags in it.

README.md
```diff
@@ -17,6 +17,96 @@ without touching any of the core components.
 
 Remember to spell mistletoe in lowercase!
 
+In This Fork
```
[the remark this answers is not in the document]
hyperking: see [commit](https://github.com/miyuchina/mistletoe/pull/170/commits/ba1fbcb1b6977457d6e51741d11b4e95ea7d2a72)

mistletoe/block_token.py
```diff
@@ -741,7 +741,6 @@ def __init__(self, content, align=None):
         self.align = align
         super().__init__(content, span_token.tokenize_inner)
 
-
```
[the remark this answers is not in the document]
hyperking: see [commit](https://github.com/miyuchina/mistletoe/pull/170/commits/6acbb2b39e0113a2f92ee85e0978b8657d44fa13)

mistletoe/block_token.py
```diff
@@ -953,6 +951,118 @@ def read(lines):
...
 
 
+class HTMLAttributes(BlockToken):
+    """
+    Block-level HTMLAttributes token.
+    This is a leaf block token without children.
+
+    Attributes:
+        raw_attr_str (str): the raw HTML attributes.
+        parent_props dict: parsed from raw_attr_str
```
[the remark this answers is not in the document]
hyperking: see [commit](https://github.com/miyuchina/mistletoe/pull/170/commits/41ae409069c9ba93ad09ff8a8d3f363f81ebf45a)

mistletoe/block_token.py
```diff
@@ -953,6 +951,118 @@ def read(lines):
...
+                if self.mapping_str in prop:
+                    key, _, value = prop.partition(self.mapping_str)
+                    return key, value
+                return None, None
+            attr_map = {}
+            for prop in attr_str.split(', '):
+                k, v = get_props(prop.strip())
+                if k and v: attr_map[k] = v
+            return attr_map
+        except Exception as e:
```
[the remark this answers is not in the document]
hyperking: see commit (https://github.com/miyuchina/mistletoe/pull/170/commits/e8256f6fa0c8aa109fef99550a6c0a82109fd553)

mistletoe/block_token.py
```diff
@@ -953,6 +951,118 @@ def read(lines):
...
+            pass
+
+    def apply_props(self, token, is_child: bool = None):
+        """Applies props recursively to parent and child tokens"""
+
+        has_nested_children = self.check_for_children(token)
+        token_props = self.parent_props if not is_child else self.child_props
+        auto_id = self.get_auto_id(token)
+        token.html_props = self.serialize(token_props, auto_id)
+        if not has_nested_children: return
```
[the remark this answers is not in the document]
hyperking: see commit (https://github.com/miyuchina/mistletoe/pull/170/commits/5cf4764284af8206bbeb96f5a36834a119a052ac)

mistletoe/block_token.py
```diff
@@ -953,6 +951,118 @@ def read(lines):
...
+        for chld in token.children:
+            is_child_key = True if chld.__class__.__name__ != 'List' else False
```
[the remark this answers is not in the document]
hyperking: see commit (https://github.com/miyuchina/mistletoe/pull/170/commits/2fac9d91144710388a6f2b59cbd8b77a27044faf)

mistletoe/block_token.py
```diff
@@ -953,6 +951,118 @@ def read(lines):
...
+            if not is_child_key:
+                self.id_index += 1
+                token_props = self.parent_props
+                child_key = token_props.get("id","uuu")
+                child_key = child_key+"-"+str(self.id_index)
```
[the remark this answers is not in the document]
hyperking: see (https://github.com/miyuchina/mistletoe/pull/170/commits/0236b95cc2636f9dd1ec9b17997462f236df913f)

mistletoe/block_token.py
```diff
@@ -953,6 +951,118 @@ def read(lines):
...
+            setattr(cls, k, v)
+
+    @classmethod
+    def get_auto_id(cls, token) -> str:
+        """Automatically generate ids for Heading elements or any specified token type"""
+        try:
+            allow_auto_id = cls.enable_auto_ids and token.__class__.__name__ in cls.allow_auto_ids
+            auto_id = token.content.lower().replace(' ','-') if allow_auto_id else ''
+            return auto_id
+        except Exception as e:
```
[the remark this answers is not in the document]
hyperking: (https://github.com/miyuchina/mistletoe/pull/170/commits/54960fc61a40e20cea1da795e76ce78f89d2c065)

test/test_html_attributes_renderer.py
```diff
@@ -0,0 +1,35 @@
...
+[some link](https://i.creativecommons.org/l/by-sa/4.0/80x15.png "toof")\
+            """
+        block_token.HTMLAttributes.configure({"enable_auto_ids": True})
+        html = markdown(txt, HTMLAttributesRenderer)
+        output = '<h1 class="foobar" id="mistletoe-is-awesome" tabindex="1">Mistletoe is Awesome</h1>\n<ul id="todos" tabindex="100">\n<li class="list-item" tabindex="1">Push Code</li>\n<li class="list-item" tabindex="1">Get Groceries\n<ul id="todos-0" tabindex="1">\n<li class="list-item" tabindex="1">Veggies</li>\n<li class="list-item" tabindex="1">Fruits\n<ul id="todos-0-1" tabindex="1">\n<li class="list-item" tabindex="1">apples</li>\n<li class="list-item" tabindex="1">oranges</li>\n</ul>\n</li>\n</ul>\n</li>\n<li class="list-item" tabindex="1">Hang up the mistletoe</li>\n</ul>\n<p class="img-sm" tabindex="1"><img src="https://i.creativecommons.org/l/by-sa/4.0/80x15.png" alt="foo" title="toof" tabindex="1" /></p>\n<p tabindex="1"><a href="https://i.creativecommons.org/l/by-sa/4.0/80x15.png" title="toof" class="btn-link" onclick="event.preventDefault();console.log(this,\'button clicked\');" tabindex="1">some link</a></p>\n'
```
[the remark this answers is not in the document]
hyperking: Hmm I don't believe the test will pass since the expected `output` would introduce new tabs characters.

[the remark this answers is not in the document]
hyperking: see (https://github.com/miyuchina/mistletoe/pull/170/commits/2fc495bdbfe661d9fa2c1fa1572a78d3df8c1cfa)

mistletoe/html_attributes_renderer.py
hyperking: See (https://github.com/miyuchina/mistletoe/pull/170/commits/e68273d0395889f17b8bfc46a49a235555320bc6)

mistletoe/html_attributes_renderer.py
hyperking: See (https://github.com/miyuchina/mistletoe/pull/170/commits/e68273d0395889f17b8bfc46a49a235555320bc6)

mistletoe/html_attributes_renderer.py
```diff
@@ -0,0 +1,256 @@
...
+        template = '<del>{}</del>'
+        return template.format(self.render_inner(token))
+
+    def render_image(self, token: span_token.Image) -> str:
+        template = '<img src="{}" alt="{}"{}{attr} />'
+        if token.title:
+            title = ' title="{}"'.format(html.escape(token.title))
+        else:
+            title = ''
+        attr = '' if not hasattr(token,'html_props') else token.html_props
```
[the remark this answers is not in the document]
hyperking: attributes are serialized ahead of time already. [See here](https://github.com/miyuchina/mistletoe/pull/170/files#diff-0d6c0ece6141f726431d55bcc65e0cd1d2edd7d05bf18531e366df030d4f81d6R1018-R1026)

features.md
```diff
@@ -0,0 +1,86 @@
+HTMLAttributesRenderer
```
anderskaplan, line 1: Nice docs! 

I'd suggest to change the name of the file to something along the lines of "Extension to attach custom html attributes" to make it more clear what it's about. "Features" is so very generic.

You could also describe the feature in a paragraph in the README.

features.md
```diff
@@ -0,0 +1,86 @@
...
+----------------------
+
+This feature allows you to write Markdown that will render 508 compliant html attributes.
```
anderskaplan, line 4: It would be nice with an explanation (maybe a link) of what "508 compliant" means. It's just a googling away but still.

mistletoe/base_renderer.py
```diff
@@ -44,6 +44,7 @@ class BaseRenderer(object):
 
     def __init__(self, *extras):
         self.render_map = {
+            'HTMLAttributes': self.render_html_attributes,
```
anderskaplan, line 47: These changes to the BaseRenderer aren't necessary and should be avoided imho. Better use the existing mechanisms for extensibility:
- The render_map entry is added automatically when you pass in HTMLAttribute into the HTMLRenderer constructor in the HTMLAttributesRenderer constructor, so you can safely remove that line.
- The render_html_attribute method can be moved to the HTMLAttributesRenderer class right away.

mistletoe/block_token.py
```diff
@@ -18,7 +18,7 @@
 """
 Tokens to be included in the parsing process, in the order specified.
 """
-__all__ = ['BlockCode', 'Heading', 'Quote', 'CodeFence', 'ThematicBreak',
+__all__ = ['HTMLAttributes', 'BlockCode', 'Heading', 'Quote', 'CodeFence', 'ThematicBreak',
```
anderskaplan, line 21: You don't need to add the HTMLAttributes class here: it's added automatically when you pass it to the HTMLRenderer constructor from the HTMLAttributesRenderer constructor.

mistletoe/block_token.py
```diff
@@ -953,6 +954,116 @@ def read(lines):
...
+        if HTMLAttributes.enable_auto_tabindex:
+            props['tabindex'] = props.get('tabindex', 1)
+        propstr = "".join([f' {k}="{v}"' for k, v in props.items()])
+        if HTMLAttributes.enable_auto_tabindex: del props['tabindex']
+        return propstr
+            
+    @classmethod
+    def configure(cls, options: dict) -> str:
+        """Override default class configuration fields"""
+        only_fields = ("start_str", "end_str", "parent_child_partition_str", "mapping_st", "allow_auto_ids", "enable_auto_ids", "enable_auto_tabindex")
```
anderskaplan, line 1031: "mapping_st": typo?

mistletoe/html_attributes_renderer.py
```diff
@@ -0,0 +1,201 @@
...
+from mistletoe.html_renderer import HTMLRenderer
+
+
+class HTMLAttributesRenderer(HTMLRenderer):
+    """
+    HTML Attributes renderer class.
+
+    See mistletoe.html_renderer module for more info.
+    """
+    def __init__(self):
```
anderskaplan, line 18: Suggest to take an "extras" parameter like the HTMLRenderer does, so that this renderer too can be extended.

mistletoe/html_attributes_renderer.py
```diff
@@ -0,0 +1,201 @@
...
+                htmlAttributesToken = token_type
+                continue
+            if htmlAttributesToken: 
+                htmlAttributesToken.apply_props(token_type)
+                htmlAttributesToken.clear()
+                htmlAttributesToken = None
+            recon_tokens.append(token_type)
+        doc_token.children = recon_tokens
+
+    def render_strong(self, token: span_token.Strong) -> str:
```
anderskaplan, line 57: This looks identical to the base class implementation (HTMLRenderer). Maybe I'm missing something, but if that's the case then I'd suggest to remove it from here.

test/test_html_attributes_renderer.py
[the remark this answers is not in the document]
anderskaplan: It would be nice with some unit tests for the HTMLAttribute options as well.

mistletoe/block_token.py
pbodnar: Yeah, I think this line is already linked / discussed in an older thread talking about performance of existing parsing and conditionality of this feature.

features.md
hyperking: See (48c8134477d1cc829a54e221d767635c8dbdb673)

features.md
hyperking: Perhaps we rename the file name from `features.md` into `extensions.md`
and within `extensions.md` we link to any available docs for new extensions or just add extension details to a single file?

mistletoe/base_renderer.py
hyperking: See (cd6ff447f92ea39b5972ef08bb81cdd21a0dc1ff)

mistletoe/block_token.py
hyperking: See (cd6ff447f92ea39b5972ef08bb81cdd21a0dc1ff)

mistletoe/block_token.py
hyperking: See (43bf8edadb96fa9805d5558e19295d5afbbdfb16)

mistletoe/html_attributes_renderer.py
hyperking: See (c6a9ed2a70268a73adb5ce4d24d567418b2e2efa)

mistletoe/html_attributes_renderer.py
hyperking: See (a7828be24ef1dba1cad7426d25561286c3be95e1)

mistletoe/base_renderer.py
```diff
@@ -143,6 +143,9 @@ def render_raw_text(self, token) -> str:
         """
         return token.content
 
+    def render_html_attributes(self, token: block_token) -> str:
```
pbodnar, line 146: After moving the new stuff out of the "base" code, also move this method to `HTMLAttributesRenderer()`, amend its type hint and return `''` from it?

pbodnar (commented):
Hi guys, I'm sorry, thanks for your efforts, but I don't really have much time for mistletoe nowadays.

So I have tried to give at least some new feedback today. I haven't gone through _all_ the code in detail yet though...

test/test_html_renderer.py
```diff
@@ -154,4 +154,4 @@ def test_footnote_link(self):
         from mistletoe import Document
         token = Document(['[name][foo]\n', '\n', '[foo]: target\n'])
         output = '<p><a href="target">name</a></p>\n' 
-        self.assertEqual(self.renderer.render(token), output)
+        self.assertEqual(self.renderer.render(token), output)
```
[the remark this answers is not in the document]
anderskaplan: (It's probably an editor setting.)

mistletoe/block_token.py
```diff
@@ -1020,6 +1020,111 @@ def read(lines):
         return [next(lines)]
 
 
+class HTMLAttributes(BlockToken):
```
pbodnar, line 1023: In the meantime, we switched to "strict/simple camel case convention" globally, so it needs to be like this now:

```suggestion
class HtmlAttributes(BlockToken):
```

The same applies to the new `HTMLAttributesRenderer`.

mistletoe/base_renderer.py
```diff
@@ -202,4 +202,4 @@ def render_thematic_break(self, token: block_token.ThematicBreak) -> str:
         return self.render_inner(token)
 
     def render_document(self, token: block_token.Document) -> str:
-        return self.render_inner(token)
+        return self.render_inner(token)
```
pbodnar, line 205: Beware of leaving EOL at the EOF everywhere.

mistletoe/block_token.py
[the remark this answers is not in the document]
pbodnar: OK, thanks.

mistletoe/html_attributes_renderer.py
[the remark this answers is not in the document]
pbodnar: Yes, although I would keep change the newly introduced method to (i.e. keep it "private"; it seems to be used by inline tokens as well; proper formatting; also move it ):

```python
    def _render_html_attributes(self, token: token.Token) -> str:
        return '' if not hasattr(token, 'html_props') else token.html_props
```

mistletoe/base_renderer.py
pbodnar: OK.

mistletoe/html_attributes_renderer.py
pbodnar: OK. Leaving this thread opened, so that we don't forget there are still some copied methods to be altered.

BTW when thinking about it, this methods' duplication is not ideal. Because when we change something in the `HtmlRenderer`, we should most probably also apply the same to the new `HtmlAttributesRenderer`, right? So putting this new feature introduced by this PR into the `HtmlRenderer` _directly_ should be our final goal, shouldn't it?

pbodnar (commented):
Still quite a lot of work on this.😅 

Also, while merging changes from the master branch is welcome, I would ask you to do Git rebase instead. I think the commit history would deserve doing some cleanup anyway, so that the current 35 commits get transformed to a few atomic commits at maximum...

PR #171, opened 2022-12-27
Merge block_token._root_node and span_token._root_node into a single object

anderskaplan:
This is a proposed refactoring to clean up the state management during parsing a tiny little bit. Instead of having separate _root_node references in the block_token and span_token modules, we merge the two into one and move it to a module already referenced by both of them: token.py.


pbodnar:
Nice DRYing, thank you. :)

PR #179, opened 2023-04-20
weird hack in PygmentsRenderer

rabinniroula:
the type mistletoe.block_token.BlockCode cannot directly access its attribute content, so I had to do this. :)

### DIFF
--- a/mistletoe/contrib/pygments_renderer.py
+++ b/mistletoe/contrib/pygments_renderer.py
@@ -14,7 +14,7 @@ class PygmentsRenderer(HTMLRenderer):
 
 
     def render_block_code(self, token):
-        code = token.content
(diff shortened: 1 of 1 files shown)


anderskaplan:
I'm not too familiar with pygments, but it's clear to see that unit tests are sorely missing. I'd suggest to add at least one test case so that this type of problems wouldn't go undetected.

Are you sure this hack is needed with the master version, though? A `content` property was added to the `BlockCode` class a while back which is supposed to do this de-referencing for you.

pbodnar:
close: I can confirm @anderskaplan's words, i.e. there seems to be no problem accessing the `BlockCode`'s property `content` since e853a0def73f4d4.

Issue #176, opened 2023-01-10
Don't convert the quotes


barthelemypousset:
Hello,

I would simply want the quotes not converted when using mistletoe:
Here's what happen
```
>>> mistletoe.markdown('"test"')
'<p>&quot;test&quot;</p>\n'
```
And I would like the result to simply be `'<p>"test"</p>\n'`
Do you know how I could do that ?

Thank you

pbodnar:
@barthelemypousset, there's no easy way how to change this in one place currently, you would have to change this in concrete calls to `html.escape()` within the `HTMLRenderer` class.

mistletoe historically escapes quotes (`"`) and since version 0.9.0 we also escape apostrophes (`'`; see #115 and ef9bd3afc1e69b23b9238b2eeee45adec627adc7). But I can see this doesn't always have to be desired, so I think it would be cool to have the escaping as an option (or options) in a future version of mistletoe. Now 1.0.0 is about to be released, so I would plan this to 1.1.0.

Cc @anderskaplan. :)

pbodnar:
@barthelemypousset (and @anderskaplan), something like in the created PR draft?

Regarding making the escaping of both double and single quotes _disabled by default_ (not for attributes though, of course!), I'm not personally against that. I always prefer simpler things if there is no evident con (and `"something"` is definitely more readable and shorter than `&quot;something&quot;`). It would just require I hope a simple change to the commonpark.py test file.

pbodnar:
> ... It would just require I hope a simple change to the commonmark.py test file, so that tests still pass.

So I've got this prepared, but just for sure, I will probably disable the double quotes escaping in the next release version after 1.1.0. Here is the diff (which would follow after the linked PR):

```diff
diff --git a/test/specification/commonmark.py b/test/specification/commonmark.py
index f3adda2..3f17c03 100644
--- a/test/specification/commonmark.py
+++ b/test/specification/commonmark.py
@@ -1,7 +1,7 @@
 import re
 import sys
 import json
-from mistletoe import markdown
+from mistletoe import Document, HTMLRenderer
 from traceback import print_tb
 from argparse import ArgumentParser

@@ -34,7 +34,8 @@ def run_tests(test_entries, start=None, end=None,
 def run_test(test_entry, quiet=False):
     test_case = test_entry['markdown'].splitlines(keepends=True)
     try:
-        output = markdown(test_case)
+        with HTMLRenderer(html_escape_double_quotes=True) as renderer:
[...]

PR #180, opened 2023-06-02
feat: optional html quotes escaping for HTMLRenderer (#176)

pbodnar:
Introduce additional parameters for the `HTMLRenderer` constructor that control double and single quotes escaping.
Also revert to previous behavior when single quotes are not escaped (by default).

### DIFF
--- a/mistletoe/contrib/mathjax.py
+++ b/mistletoe/contrib/mathjax.py
@@ -6,10 +6,14 @@ from mistletoe.html_renderer import HTMLRenderer
 from mistletoe.latex_renderer import LaTeXRenderer
 
 class MathJaxRenderer(HTMLRenderer, LaTeXRenderer):
-    """
-    MRO will first look for render functions under HTMLRenderer,
-    then LaTeXRenderer.
-    """
+    def __init__(self, **kwargs):
+        """
+        Args:
+            **kwargs: additional parameters to be passed to the ancestors'
+                      constructors.
+        """
+        super().__init__(**kwargs)
+
     mathjax_src = '<script src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-MML-AM_CHTML"></script>\n'
 
     def render_math(self, token):
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -18,7 +18,7 @@ class HTMLRenderer(BaseRenderer):
 
     See mistletoe.base_renderer module for more info.
     """
-    def __init__(self, *extras, html_escape_double_quotes=True, html_escape_single_quotes=False):
+    def __init__(self, *extras, html_escape_double_quotes=True, html_escape_single_quotes=False, **kwargs):
         """
         Args:
             extras (list): allows subclasses to add even more custom tokens.
@@ -26,9 +26,11 @@ class HTMLRenderer(BaseRenderer):
                 quotes when HTML-escaping rendered text.
             html_escape_single_quotes (bool): whether to also escape single
                 quotes when HTML-escaping rendered text.
+            **kwargs: additional parameters to be passed to the ancestor's
+                constructor.
         """
         self._suppress_ptag_stack = [False]
-        super().__init__(*chain((HTMLBlock, HTMLSpan), extras))
+        super().__init__(*chain((HTMLBlock, HTMLSpan), extras), **kwargs)
         self.html_escape_double_quotes = html_escape_double_quotes
         self.html_escape_single_quotes = html_escape_single_quotes
 
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -18,13 +18,19 @@ class HTMLRenderer(BaseRenderer):
 
     See mistletoe.base_renderer module for more info.
     """
-    def __init__(self, *extras):
+    def __init__(self, *extras, html_escape_double_quotes=True, html_escape_single_quotes=False):
         """
         Args:
             extras (list): allows subclasses to add even more custom tokens.
+            html_escape_double_quotes (bool): whether to also escape double
+                quotes when HTML-escaping rendered text.
+            html_escape_single_quotes (bool): whether to also escape single
+                quotes when HTML-escaping rendered text.
         """
         self._suppress_ptag_stack = [False]
         super().__init__(*chain((HTMLBlock, HTMLSpan), extras))
+        self.html_escape_double_quotes = html_escape_double_quotes
+        self.html_escape_single_quotes = html_escape_single_quotes
 
     def __exit__(self, *args):
         super().__exit__(*args)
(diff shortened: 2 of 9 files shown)


mistletoe/html_renderer.py
```diff
@@ -193,12 +199,20 @@ def render_document(self, token: block_token.Document) -> str:
...
         """
-        This method is deprecated. Use `html.escape` instead.
+        Like `html.escape()`, but this  looks into the current rendering options
+        to decide which of the quotes (double, single, or both) to escape.
+
+        Intended for escaping text content. To escape content of an attribute,
+        simply call `html.escape()`.
         """
-        return html.escape(raw)
+        s = html.escape(s, quote=False)
```
pbodnar, line 210: As `html.escape` doesn't seem to be a native function, in-lining it here should be faster?

mistletoe/contrib/mathjax.py
```diff
@@ -6,10 +6,17 @@
 from mistletoe.latex_renderer import LaTeXRenderer
 
 class MathJaxRenderer(HTMLRenderer, LaTeXRenderer):
-    """
-    MRO will first look for render functions under HTMLRenderer,
-    then LaTeXRenderer.
-    """
+    def __init__(self, **kwargs):
+        """
+        MRO will first look for render functions under HTMLRenderer,
```
anderskaplan, line 11: Maybe "MRO" deserves an explanation?

mistletoe/html_renderer.py
```diff
@@ -18,13 +18,19 @@ class HTMLRenderer(BaseRenderer):
 
     See mistletoe.base_renderer module for more info.
     """
-    def __init__(self, *extras):
+    def __init__(self, *extras, html_escape_double_quotes=True, html_escape_single_quotes=False):
```
anderskaplan, line 21: "html-escape" or just "escape"? Imho, escape is sufficient.

mistletoe/html_renderer.py
```diff
@@ -193,12 +199,20 @@ def render_document(self, token: block_token.Document) -> str:
...
+
+        Intended for escaping text content. To escape content of an attribute,
+        simply call `html.escape()`.
         """
-        return html.escape(raw)
+        s = html.escape(s, quote=False)
+        if self.html_escape_double_quotes:
+            s = s.replace('"', "&quot;")
+        if self.html_escape_single_quotes:
+            s = s.replace('\'', "&#x27;")
```
anderskaplan, line 214: Maybe `&apos;` instead of the ascii value, similar to the double quote?

mistletoe/html_renderer.py
anderskaplan: An advantage of inlining it would be to avoid the clever but scary patching of `html._charref` done in `span_tokenizer.tokenize()`.

mistletoe/html_renderer.py
anderskaplan: Ah now I see this is how it's done in `html.escape()`. So there's probably a good reason for it. Never mind.

mistletoe/html_renderer.py
anderskaplan, line 21: Should the `HTMLRenderer.init()` perhaps also take **kwargs parameters? And `Scheme`, too?

test/specification/commonmark.py
```diff
@@ -34,7 +34,7 @@ def run_tests(test_entries, start=None, end=None,
 def run_test(test_entry, quiet=False):
     test_case = test_entry['markdown'].splitlines(keepends=True)
     try:
-        output = markdown(test_case).replace('&#x27;', "'")
```
anderskaplan, line 37: Does this mean that the CommonMark tests now pass without special handling of escaped quotes? If yes, then that's a big win!

test/test_html_renderer.py
```diff
@@ -138,6 +138,19 @@ def test_document(self):
         self._test_token('Document', '', footnotes={})
 
 
+class TestHTMLRendererEscaping(TestCase):
+    @parameterized.expand([
```
anderskaplan, line 142: Nice way to put it.

mistletoe/html_renderer.py
anderskaplan: Turns out [`&apos;` is an XML entity reference but not HTML](https://fishbowl.pastiche.org/2003/07/01/the_curse_of_apos). Odd.

mistletoe/contrib/mathjax.py
pbodnar: Yeah, maybe, it is <https://docs.python.org/3/library/stdtypes.html#class.__mro__>. But I didn't want to dig any deeper for now. :P

mistletoe/html_renderer.py
pbodnar: I was asking myself the same question. :) And I decided to go with the longer variant, so I can be sure this won't collide for example with `latex_escape_double_quotes` or similar in the future, no matter how improbable this is now. ;)

mistletoe/html_renderer.py
pbodnar: `html._charref` is used for `html.unescape` though? :)

test/specification/commonmark.py
pbodnar: Yes. I would say a little win here. ;)

test/test_html_renderer.py
pbodnar: Thanks. :)

pbodnar (commented):
@anderskaplan, thanks for the review! 👍

mistletoe/html_renderer.py
anderskaplan: Yes, you're right. It won't help for that.

mistletoe/contrib/mathjax.py
anderskaplan: How about changing "MRO" to "Python", then? It would make the comment understandable without digging into the technicalities.

anderskaplan: (It is off-topic for this PR, for sure. But still could make sense, as the line is touched.)

pbodnar: You;re right, we are moving it anyway. It is rather an inner workings note, explaining what Python does when there are multiple ancestors, so I would remove it altogether, what do you say?

anderskaplan: Yes, sounds good!

pbodnar: Done. :)

mistletoe/html_renderer.py
pbodnar: Done.

mistletoe/html_renderer.py
pbodnar: FYI, I in-lined the code in the end. Not that it would be visibly faster, but I guess it's the way to go in this case.

mistletoe/html_renderer.py
pbodnar: Oddly enough, I can see now that I have somehow overlooked to also extend the constructor of `PygmentsRenderer` - will do additionally...

PR #183, opened 2023-06-17
fixing "pygments.util.ClassNotFound: no lexer for alias" error

derekn:
this fixes an error when an unknown language is encountered in the PygmentsRenderer

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -32,7 +32,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install flake8 pytest parameterized
+        python -m pip install flake8 pytest parameterized pygments
         if [ -f requirements.txt ]; then python -m pip install -r requirements.txt; fi
     - name: Lint with flake8
       run: |
@@ -65,7 +65,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install pytest parameterized
+        python -m pip install pytest parameterized pygments
         # note: the following also installs "coverage"
         python -m pip install coveralls
         if [ -f requirements.txt ]; then python -m pip install -r requirements.txt; fi
--- a/mistletoe/contrib/pygments_renderer.py
+++ b/mistletoe/contrib/pygments_renderer.py
@@ -1,20 +1,32 @@
 from mistletoe import HTMLRenderer
 from pygments import highlight
-from pygments.styles import get_style_by_name as get_style
-from pygments.lexers import get_lexer_by_name as get_lexer, guess_lexer
 from pygments.formatters.html import HtmlFormatter
+from pygments.lexers import get_lexer_by_name as get_lexer, guess_lexer
+from pygments.styles import get_style_by_name as get_style
+from pygments.util import ClassNotFound
+
 
 class PygmentsRenderer(HTMLRenderer):
     formatter = HtmlFormatter()
     formatter.noclasses = True
 
-    def __init__(self, *extras, style='default'):
+    def __init__(self, *extras, style='default', fail_on_unsupported_language=False):
         super().__init__(*extras)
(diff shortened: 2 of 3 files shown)


pbodnar:
@derekn, thank you, LGTM. Do you think you could also add unit tests which would cover the fix (no language vs. known language vs. unknown language)? Looks like `test_pygments_renderer.py` is missing, so it needs to be created from scratch...

pbodnar:
Also, I think we should give clients the possibility to still get an error when an unknown language is passed. I.e. introduce a new constructor parameter, say `fail_on_unknown_language=True` (`True` for backwards compatibility), and re-throw the `ClassNotFound` exception if this parameter is `True` (and `token.language` is not empty).

derekn:
Added unit tests for PygmentsRenderer.

Also, added the `fail_on_unknown_language` option to the renderer, but defaulted it to False. I believe this aligns with the behavior of other markdown packages and is generally more desirable in this situation. I dont think this will be a breaking change for clients.

pbodnar:
@derekn, good point, I agree.

Just a small correction of the handling of empty language supplied (I'm not sure if I want to commit into your _master_ branch - it is usually better to create a dedicated branch for PR, e.g. `fix-pygments-classnotfound` in this case):

```diff
diff --git a/mistletoe/contrib/pygments_renderer.py b/mistletoe/contrib/pygments_renderer.py
index c07a39c..0d3bd29 100644
--- a/mistletoe/contrib/pygments_renderer.py
+++ b/mistletoe/contrib/pygments_renderer.py
@@ -18,11 +18,14 @@ class PygmentsRenderer(HTMLRenderer):
     def render_block_code(self, token):
         code = token.content

-        try:
-            lexer = get_lexer(token.language)
-        except ClassNotFound as err:
-            if self.fail_on_unknown_language:
-                raise err
+        lexer = None
+        if token.language:
+            try:
+                lexer = get_lexer(token.language)
+            except ClassNotFound as err:
+                if self.fail_on_unknown_language:
+                    raise err
+        if lexer is None:
             lexer = guess_lexer(code)

         return highlight(code, lexer, self.formatter)
diff --git a/test/test_contrib/test_pygments_renderer.py b/test/test_contrib/test_pygments_renderer.py
index 491c827..839ea55 100644
--- a/test/test_contrib/test_pygments_renderer.py
+++ b/test/test_contrib/test_pygments_renderer.py
@@ -2,12 +2,17 @@ import unittest

 from mistletoe import Document
 from mistletoe.contrib.pygments_renderer import PygmentsRenderer
+from parameterized import parameterized
 from pygments.util import ClassNotFound


 class TestPygmentsRenderer(unittest.TestCase):
-    def test_render_no_language(self):
-        renderer = PygmentsRenderer()
+    @parameterized.expand([
+        (True),
+        (False)
+    ])
+    def test_render_no_language(self, fail_on_unknown_language: bool):
+        renderer = PygmentsRenderer(fail_on_unknown_language=fail_on_unknown_language)
         token = Document(['```\n', 'no language\n', '```\n'])
         output = renderer.render(token)
         actual = (
```

In relation to this fix, I've realized we should better name the new option `fail_on_unsupported_language` - can you do the rename and squash the changes to your last commit? Thank you.

derekn:
@pbodnar added your changes, renamed `fail_on_unsupported_language` option, and squashed

derekn:
Added the pygments dependency to the github action for the tests

pbodnar:
Damn it, one more thing it seems - older Python build say this:

> TypeError: Parameters must be tuples, but True is not (hint: use '(True, )')

pbodnar:
> Added the pygments dependency to the github action for the tests

Thank you, you are quick. :) I will probably try to introduce [requirements.txt](https://pip.pypa.io/en/latest/user_guide/#requirements-files) in order to avoid duplication, but in a later separate commit.

pbodnar:
@derekn, good job, thank you. :)

derekn:
@pbodnar thanks, happy to help!

PR #185, opened 2023-06-21
test: unify assertEqual calls in all the tests

pbodnar:
It's easy and unified now (with no exceptions, right?! :)):

1. As per unittest's documentation examples, `actual` always goes before `expected`.
2. If used, `actual` is stored in a variable `output` for us.
3. If used, variable `expected` is defined after the `output` variable.

For curious ones, Python (as well as the rest of the world) is not consistent here: https://stackoverflow.com/questions/40945126/parameter-order-for-unittest-testcase-assertequal.


pbodnar:
@anderskaplan, this might be of an interest for you. :)

anderskaplan:
Looks good to me! 👍

Issue #166, opened 2022-11-18
Enable tables which interrupt a paragraph (like GFM does)


minaminao:
Input markdown:
```
A:
|a|b|
|-|-|
|a|b|
```

`ast_renderer` outputs:
```
    {
      "type": "Paragraph",
      "children": [
        {
          "type": "RawText",
          "content": "A:"
        },
        {
          "type": "LineBreak",
          "soft": true,
          "content": ""
        },
        {
          "type": "RawText",
          "content": "|a|b|"
        },
        {
          "type": "LineBreak",
          "soft": true,
          "content": ""
        },
        {
          "type": "RawText",
          "content": "|-|-|"
        },
        {
          "type": "LineBreak",
          "soft": true,
          "content": ""
        },
        {
          "type": "RawText",
          "content": "|a|b|"
        }
      ]
    },
```

I believe `Table` should be output, not `RawText`.

#### Actual output in GitHub:
A:
|a|b|
|-|-|
|a|b|

pbodnar:
@minaminao, thanks for the report. After some analysis, I would consider this as a change request which could be stated like this:

_Enable tables to interrupt a paragraph_

Let me explain why:

1. Table like this works fine:

pbodnar:
FYI, by mistake, I have submitted the previous comment before finishing it. It is ready now. :)

minaminao:
Thanks for the analysis :)

(I found this in a table with more than 3 consecutive `-`'s, but simplified the table for reporting, and then have related it to another isssue #131 unexpectedly)

anderskaplan:
As mistletoe has the ambition to support custom tokens, it would be nice if it also had a mechanism to customize which tokens are allowed to interrupt a paragraph. As you point out @pbodnar , this is currently hard coded in `Paragraph.read()`. With such a mechanism in place, it should be straightforward to enable the desired behavior for tables.

I have some ideas about how to do this, so I'll try and make a PR out of them.

pbodnar:
@anderskaplan, that looks promising - having it implemented in a more generic way, not just for `Table`, would be great. :)

anderskaplan:
Created two PR's: one "infrastructure" fix (#186) and a draft PR for the new mechanism + table change (#187).

PR #186, opened 2023-06-24
Changed FileWrapper anchor/reset to getpos/setpos.

anderskaplan:
Issue found during implementation of #166, when checking for the start of `Table` tokens.

When checking if a table starts on a line, one needs to scan multiple lines without consuming any content. The way to do this is with the anchor/reset methods on the `FileWrapper` class. However, this breaks the parsing of `List` tokens, because these also use anchor/reset, and the anchor set by `List` gets overwritten by the one set in `Table`.

The proposed solution is to not store the anchor position on the `FileWrapper`, but instead move that responsibility to the parsing method. This way, recursive parsing can hold multiple anchors at the same time, without any interference.

### DIFF
--- a/mistletoe/block_tokenizer.py
+++ b/mistletoe/block_tokenizer.py
@@ -7,7 +7,6 @@ class FileWrapper:
     def __init__(self, lines):
         self.lines = lines if isinstance(lines, list) else list(lines)
         self._index = -1
-        self._anchor = 0
 
     def __next__(self):
         if self._index + 1 < len(self.lines):
@@ -21,11 +20,11 @@ class FileWrapper:
     def __repr__(self):
         return repr(self.lines[self._index+1:])
 
-    def anchor(self):
-        self._anchor = self._index
+    def getpos(self):
+        return self._index
 
-    def reset(self):
-        self._index = self._anchor
+    def setpos(self, pos):
+        self._index = pos
 
     def peek(self):
         if self._index + 1 < len(self.lines):
--- a/mistletoe/block_tokenizer.py
+++ b/mistletoe/block_tokenizer.py
@@ -7,6 +7,7 @@ class FileWrapper:
     def __init__(self, lines):
         self.lines = lines if isinstance(lines, list) else list(lines)
         self._index = -1
+        self._anchor = 0
 
     def __next__(self):
         if self._index + 1 < len(self.lines):
@@ -29,6 +30,13 @@ class FileWrapper:
         """Sets the current reading position."""
         self._index = pos
 
+    def anchor(self):
+        """@deprecated use `get_pos` instead"""
+        self._anchor = self.get_pos()
+
+    def reset(self):
+        """@deprecated use `set_pos` instead"""
+        self.set_pos(self._anchor)
     def peek(self):
         if self._index + 1 < len(self.lines):
             return self.lines[self._index+1]
--- a/mistletoe/block_tokenizer.py
+++ b/mistletoe/block_tokenizer.py
@@ -20,10 +20,13 @@ class FileWrapper:
     def __repr__(self):
         return repr(self.lines[self._index+1:])
 
-    def getpos(self):
+    def get_pos(self):
+        """Returns the current reading position.
+        The result is an opaque value which can be passed to `set_pos`."""
         return self._index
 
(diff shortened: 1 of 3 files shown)


anderskaplan:
@pbodnar Please disregard the test coverage check, it's a false alarm caused by deleting one line of "tested" code.

mistletoe/block_tokenizer.py
```diff
@@ -21,11 +20,11 @@ def __iter__(self):
...
 
-    def anchor(self):
-        self._anchor = self._index
+    def getpos(self):
+        return self._index
 
-    def reset(self):
-        self._index = self._anchor
+    def setpos(self, pos):
+        self._index = pos
```
pbodnar, line 27: Can we please rename these to more standard `get_pos` and `set_pos`?

Or, I guess we could alternatively turn `pos` into [property](https://realpython.com/python-getter-setter/#using-properties-instead-of-getters-and-setters-the-python-way) - nice-to-have? :)

pbodnar (changes requested):
@anderskaplan, good catch. :) Just one suggestion to the new methods naming...

pbodnar:
Another thought: What about instead of replacing the existing anchor/reset methods making them _stacked_ (a simple list for `_index` would probably suffice)? Would that make sense to you?

Also, we should expect that some could use the existing methods in their custom tokens, so that is another reason for not removing them right away.

mistletoe/block_tokenizer.py
anderskaplan: Sure, I can rename to get_pos and set_pos if that's more standard. I think that's better than using a property, because the setter would kind-of have side effects.

anderskaplan:
> Another thought: What about instead of replacing the existing anchor/reset methods making them _stacked_ (a simple list would probably suffice for the stored anchors)? Would that make sense to you?

Yes, I did consider that option, I even tried implementing it :) However, that option assumes that a push is always followed by a pop, and that is not the case here. Sure, one could force the calling code to clean up (i.e. pop) after itself, but that becomes more complex and error prone than just letting the calling code keep track of the stored position.

> Also, we should expect that some could use the existing methods in their custom tokens, so that is another reason for not removing them right away.

It could happen, but I'd guess that the risk is rather small. Imho this is one of those band-aids that just need to be pulled. We *could* keep the anchor/reset methods for now, though, and mark them as deprecated. Which do you prefer?

mistletoe/block_tokenizer.py
pbodnar: Thank you for the renaming.

> ... because the setter would kind-of have side effects

Out of curiosity, what side effects do you mean, besides storing the value?

pbodnar:
> > Another thought: What about instead of replacing the existing anchor/reset methods making them _stacked_ (a simple list would probably suffice for the stored anchors)? Would that make sense to you?
> 
> Yes, I did consider that option, I even tried implementing it :) However, that option assumes that a push is always followed by a pop, and that is not the case here. Sure, one could force the calling code to clean up (i.e. pop) after itself, but that becomes more complex and error prone than just letting the calling code keep track of the stored position.

OK, I would expect that push and pop would by always stacked. But anyway yes, it is definitely easier this way, even if we need to introduce a new variable (`anchor`) in every such block.

> > Also, we should expect that some could use the existing methods in their custom tokens, so that is another reason for not removing them right away.
> 
> It could happen, but I'd guess that the risk is rather small. Imho this is one of those band-aids that just need to be pulled. We _could_ keep the anchor/reset methods for now, though, and mark them as deprecated. Which do you prefer?

I prefer to keep them and mark them as deprecated, please. :)

mistletoe/block_tokenizer.py
anderskaplan: The side effect I was thinking of is that the next time you read a line, it will be from somewhere else in the buffer.

pbodnar: Thanks for the clarification. So your point is probably that it might happen that some users of `lines.pos = X` could expect something else from this call than when using `lines.set_pos(X)`? I personally don't think so, but never mind, let's move on. :)

pbodnar (approved):
Fine from me, thanks for also adding the tests.

PR #187, opened 2023-06-25
Make tables interrupt paragraphs (#166)

anderskaplan:
This PR introduces a more modular way to detect when a block token interrupts a paragraph. It also makes `Table` tokens interrupt paragraphs, using the new mechanism.

Adding it as a draft PR for now, due to the dependency on #186 .

In addition to the "obvious" use in `Paragraph.read()`, checks for paragraph interruptions are also done in `Quote.read()` and `ListItem.read()` to determine whether a line is a lazy continuation line or not.

Using the new mechanism to check for continuation lines in `Quote` and `ListItem` does change the parsing behavior, and I believe it's for the better. The old implementation of these checks did not include `HTMLBlock`, for example. There are no test cases in the CommonMark test suite to prove it, but I believe the new implementation is closer to the definition of "[Paragraph continuation text](https://spec.commonmark.org/0.30/#paragraph-continuation-text)" in the CommonMark spec.

Up for discussion: as it says in the issue report (#166), making `Table` interrupt paragraphs is a change of behavior, and maybe it should be a setting. If so, I'd propose to add the setting to the `Table` class, rather than to make it a renderer parameter, because this is a parsing option and it has nothing to do with the rendering.

### DIFF
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -93,7 +93,7 @@ class BlockToken(token.Token):
           that should be read into this token.
 
           Default to stop at an empty line.
-          
+
           Note that BlockToken.read does not have to return a list of lines.
           Because the return value of this function will be directly
           passed into the token constructor, we can return any relevant
@@ -113,7 +113,7 @@ class BlockToken(token.Token):
 
     def __contains__(self, text):
         return any(text in child for child in self.children)
-    
+
     @staticmethod
     def read(lines):
         line_buffer = [next(lines)]
@@ -173,9 +173,14 @@ class Heading(BlockToken):
         return True
 
     @classmethod
+    def check_interrupts_paragraph(cls, line):
+        return cls.start(line)
+
+    @classmethod
     def read(cls, lines):
         next(lines)
         return cls.level, cls.content, cls.closing_sequence
+
 
 class SetextHeading(BlockToken):
     """
@@ -219,6 +224,10 @@ class Quote(BlockToken):
         if len(line) - len(stripped) > 3:
             return False
         return stripped.startswith('>')
+
+    @classmethod
+    def check_interrupts_paragraph(cls, line):
+        return cls.start(line)
 
     @classmethod
     def read(cls, lines):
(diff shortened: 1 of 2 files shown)


pbodnar:
@anderskaplan, while rebasing, I think you would also want to restructure the commits a bit - namely the very last commit currently (6ed847e07fc9203f2b1f872dcbfba51fb0db4b16) content doesn't seem to quite match what is in the commit message?

anderskaplan:
> @anderskaplan, while rebasing, I think you would also want to restructure the commits a bit - namely the very last commit currently ([6ed847e](https://github.com/miyuchina/mistletoe/commit/6ed847e07fc9203f2b1f872dcbfba51fb0db4b16)) content doesn't seem to quite match what is in the commit message?

I've clarified the commit message a little, but overall I think it was correct -- it's the override of the `check_interrupts_paragraph` method that does the trick, once the groundwork has been done.

pbodnar (changes requested):
Thank you, LGTM! :)

Maybe just what about squashing the 2 commits which follow after e2829a43575c84a750? Or changing their commit message a bit. So that the relation among them (doing de-facto the same for Paragraphs, but also for Quotes and List items) is more visible?

pbodnar:
> Up for discussion: as it says in the issue report (#166), making `Table` interrupt paragraphs is a change of behavior, and maybe it should be a setting. If so, I'd propose to add the setting to the `Table` class, rather than to make it a renderer parameter, because this is a parsing option and it has nothing to do with the rendering.

Yes, I would make this a _parsing_ option, i.e. to keep it separate from the options we pass to the various renderers now. I haven't thought about the exact form yet, so if you've already got an idea, that would be great... Maybe introducing a global "parsing options" variable would be the only way for now, without the necessity to do a major refactoring?

anderskaplan:
> Maybe just what about squashing the 2 commits which follow after [e2829a4](https://github.com/miyuchina/mistletoe/commit/e2829a43575c84a750af4b16fdf38c61ccafd7a4)? Or changing their commit message a bit. So that the relation among them (doing de-facto the same for Paragraphs, but also for Quotes and List items) is more visible?

Sure, I think even the 4 commits after [e2829a4] could be squashed into one.

anderskaplan:
> Yes, I would make this a _parsing_ option, i.e. to keep it separate from the options we pass to the various renderers now. I haven't thought about the exact form yet, so if you've already got an idea, that would be great... Maybe introducing a global "parsing options" variable would be the only way for now, without the necessity to do a major refactoring?

I'll think about it and come up with a suggestion :)

pbodnar (approved):

pbodnar:
@anderskaplan, thanks a lot, good job. 👍 

So we newly allow tables to interrupt paragraphs by default - I will address this in the release notes, mentioning the option `block_token.Table.interrupt_paragraph` there.

Issue #144, opened 2022-05-23
Record line numbers on tokens


djmattyg007:
I'm writing a tool to parse markdown files and verify that links are valid. To be able to provide the most valuable, accurate feedback, I need to be able to display line numbers to the user, so that they know exactly where in a file the broken link is.

Right now this isn't possible, because line numbers aren't recorded on token objects.

pbodnar:
@djmattyg007, I think that what you propose would be a pretty cool feature. In principal, it shouldn't be hard to implement, yet it would probably require some broader refactoring in order to track & store the context information (line, plus column would be also nice). I would also watch out for possible impact on performance, but I don't expect some human-noticeable difference...

I would like to look at this one day, but I don't have much time these days, so, as usually, PRs are welcome in the meantime. :)

anderskaplan:
Hi, I have a somewhat similar use case, where I'd like to know the position of translatable content in the Markdown document.

I've started on an implementation. It turns out that block tokens are fairly straightforward, but span tokens are more difficult because the newlines are discarded before parsing, see e.g. the Paragraph constructor.

@djmattyg007: If you could get the starting line number of the paragraph containing the link, would that be of any use to you? Or do you need the line number of the link?

djmattyg007:
I need the line number (and ideally the offset on the given line).

anderskaplan:
Ok, I have a PR in the works for line numbers on block tokens which may or may not be useful for you. But it depends on #172, so it will have to wait until that one is merged.

pbodnar:
@anderskaplan, while implementing the `MarkdownRenderer` recently, haven't you investigated this feature request more? I think it would be pretty cool and possibly another distinguishing trait of mistletoe? :)

pbodnar:
Thanks to @anderskaplan, we've got this implemented for _block_ tokens in #188. It still doesn't fully cover the original request though (finding out line number for a any, notably _span_ token), so I will leave this issue open for now.

raffaeler:
In addition to the line, the start/end offset from the beginning of the file for each block would be precious.

PR #188, opened 2023-06-25
Add line numbers on all block tokens during parsing (#144)

anderskaplan:
This is a partial solution for #144 which records line numbers on all _block_ tokens. The _span_ tokens are still left unlabeled, but at least it's a starting point, and I think the functionality added here is useful by itself.

Open question: should the `line_number` be a added to `repr_attributes`, so that it gets included in the output from `ASTRenderer`?

### DIFF
--- a/dev-guide.md
+++ b/dev-guide.md
@@ -3,13 +3,25 @@
 This document describes usage of mistletoe and its API
 from the developer's point of view.
 
-Understanding the AST
----------------------
+Understanding the AST and the tokens
+------------------------------------
 
 When a markdown document gets parsed by mistletoe, the result is represented
-as an "abstract syntax tree" (AST), stored in an instance of `Document`.
-This object contains a hierarchy of
-all the various tokens which were recognized during the parsing process.
+as an _abstract syntax tree (AST)_, stored in an instance of `Document`.
+This object contains a hierarchy of all the various _tokens_ which were recognized
+during the parsing process, for example, `Paragraph`, `Heading`, and `RawText`.
+
+The tokens which represent a line or a block of lines in the input markdown
+are called _block tokens_. Examples include `List`, `Paragraph`, `ThematicBreak`,
+and also the `Document` itself.
+
+The tokens which represent the actual content within a block are called _span tokens_,
+or, with CommonMark terminology, _inline tokens_.
+In this category you will find tokens like `RawText`, `Link`, and `Emphasis`.
+
+Block tokens may have block tokens, span tokens, or no tokens at all as children
+in the AST; this depends on the type of token. Span tokens may *only* have span
+tokens as children.
 
 In order to see what exactly gets parsed, one can simply use the `AstRenderer`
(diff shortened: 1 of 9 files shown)


pbodnar:
@anderskaplan, thank you, I will try to look at this for the next release.

pbodnar:
Finally, I have found time for mistletoe once again. :)

@anderskaplan, I must say "good job" 👍 . Basically, I like it, including the clever way of the new unit test.

Probably just 3 more things be resolved:

1. Performance - did you run the benchmark test? I guess there should be little to zero difference to the version without line numbers, right?
2. Optionality - shouldn't we better pass the `line_number` parameter as the last AND optional (with default value of `None`) to the constructors of `TableRow` etc. (even `Table`?)? This way, we could achieve backwards compatibility for those users who might create (some) tokens without actually parsing a markdown text. Yes, there probably won't be many, but what if there are some out there? :)
3. Documentation - probably add some notes on this cool feature as a sub-chapter to https://github.com/miyuchina/mistletoe/blob/v1.2.1/dev-guide.md#understanding-the-ast?

> Open question: should the `line_number` be a added to `repr_attributes`, so that it gets included in the output from `ASTRenderer`?

Not sure here, but probably yes, it could? :)

anderskaplan:
Hey @pbodnar, just checking in to say hi. The last weeks have been quite busy with other things but I'll get back to this as soon as I can.

anderskaplan:
I have updated the documentation, made the line_number parameter optional, and run the benchmark as requested in the review.

The benchmark results have a fairly large variance (5 %CV) from run to run on my machine. But I'm fairly confident that the performance penalty of the line numbers is negligible, because if I compare the mean benchmark with and without line numbers from five runs of each type, then the relative difference is about 0.1%. I.e., far less than the run-to-run variance.

I also added the `line_number` attribute to `repr_attributes`, so that the line numbers are included in the output from the ASTRenderer.

dev-guide.md
```diff
@@ -86,12 +103,25 @@ Then we will get this JSON output from the AST renderer:
 }
 ```
 
-When passing this tree to a renderer, it is recursively traversed
+### Line numbers
+
+mistletoe records the starting line of all block tokens that it encounters during
+parsing and stores it the `line_number` attribute.
```
pbodnar, line 109:
```diff
-parsing and stores it the `line_number` attribute.
+parsing and stores it as the `line_number` attribute of each token.
```

pbodnar (changes requested):
@anderskaplan, thank you for the amendments, I've found just one typo. :)

Beside that, it looks like some other conflicting commits were made in the mean time...

dev-guide.md
anderskaplan: Fixed, along with the merge conflicts :)

pbodnar: Perfect, thank you! :)

pbodnar (approved):

PR #189, opened 2023-06-25
Added a unit test for the Scheme renderer.

anderskaplan:
...because it's nice with at least _some_ test coverage.

### DIFF
--- a/mistletoe/contrib/scheme.py
+++ b/mistletoe/contrib/scheme.py
@@ -153,11 +153,3 @@ class Scheme(BaseRenderer):
         finally:
             self.env = old_env
         return result
-
-
-if __name__ == '__main__':
-    with Scheme() as renderer:
-        prog = ["(define x (* 2 21))",
-                "x"]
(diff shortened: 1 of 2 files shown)


pbodnar:
@anderskaplan, good "catch". And what do you think about removing this section from `scheme.py` then:

```python
if __name__ == '__main__':
    with Scheme() as renderer:
        prog = ["(define x (* 2 21))",
                "x"]
        print(renderer.render(Program(prog)))
```

? I think for consistency with other renderers, removing it wouldn't be bad.

anderskaplan:
> @anderskaplan, good "catch". And what do you think about removing this section from `scheme.py` then:
> 
> ```python
> if __name__ == '__main__':
>     with Scheme() as renderer:
>         prog = ["(define x (* 2 21))",
>                 "x"]
>         print(renderer.render(Program(prog)))
> ```
> 
> ? I think for consistency with other renderers, removing it wouldn't be bad.

I agree. I'll do that!

PR #190, opened 2023-06-27
Escape special LaTeX characters in URLs

choeppler:
If `\url{}` is used within macros like `\multicolumn` and if the link to
be rendered contains `#` or `%` as typical confluence links like
https://site.com/confluence/display/TEST/TEST-042%3A+Foo#target
are bound to, pdflatex (from TeX Live 2019) fails with errors. This
patch fixes that by escaping these two characters.

NB: if `\url{}` with such links is used outside of multicolumn-macros,
they work regardless of whether these characters are escaped, and the
result remains the same.

### DIFF
--- a/mistletoe/latex_renderer.py
+++ b/mistletoe/latex_renderer.py
@@ -2,11 +2,11 @@
 LaTeX renderer for mistletoe.
 """
 
+import string
 from itertools import chain
+from urllib.parse import quote
 import mistletoe.latex_token as latex_token
 from mistletoe.base_renderer import BaseRenderer
-import string
-
 
 # (customizable) delimiters for inline code
 verb_delimiters = string.punctuation + string.digits
@@ -61,11 +61,12 @@ class LaTeXRenderer(BaseRenderer):
         self.packages['hyperref'] = []
         template = '\\href{{{target}}}{{{inner}}}'
         inner = self.render_inner(token)
-        return template.format(target=token.target, inner=inner)
+        return template.format(target=self.escape_url(token.target),
+                               inner=inner)
 
     def render_auto_link(self, token):
         self.packages['hyperref'] = []
-        return '\\url{{{}}}'.format(token.target)
+        return '\\url{{{}}}'.format(self.escape_url(token.target))
 
     @staticmethod
     def render_math(token):
@@ -174,3 +175,22 @@ class LaTeXRenderer(BaseRenderer):
         self.footnotes.update(token.footnotes)
         return template.format(inner=self.render_inner(token),
                                packages=self.render_packages())
+
+    @staticmethod
+    def escape_url(raw: str) -> str:
+        """
+        Quote unsafe chars in urls & escape as needed for LaTeX's hyperref.
+
+        %-escapes all characters that are neither in the unreserved chars
+        ("always safe" as per RFC 2396 or RFC 3986) nor in the chars set
+        '/#:()*?=%@+,&;'
+
+        Subsequently, LaTeX-escapes '%' and '#' for hyperref's \\url{} to also
+        work if used within macros like \\multicolumn. if \\url{} with urls
+        containing '%' or '#' is used outside of multicolumn-macros, they work
+        regardless of whether these characters are escaped, and the result
+        remains the same (at least for pdflatex from TeX Live 2019).
+        """
+        quoted_url = quote(raw, safe='/#:()*?=%@+,&;')
+        return quoted_url.replace('%', '\\%') \
+                         .replace('#', '\\#')
--- a/test/test_latex_renderer.py
+++ b/test/test_latex_renderer.py
@@ -1,4 +1,5 @@
 from unittest import TestCase, mock
+from parameterized import parameterized
 import mistletoe.latex_renderer
 from mistletoe.latex_renderer import LaTeXRenderer
 from mistletoe import markdown
@@ -50,13 +51,21 @@ class TestLaTeXRenderer(TestCase):
         output = '\n\\includegraphics{src}\n'
         self._test_token('Image', output, src='src')
 
-    def test_link(self):
-        output = '\\href{target}{inner}'
-        self._test_token('Link', output, target='target')
+    @parameterized.expand([
+        ('page', '\\href{page}{inner}'),
+        ('page%3A+with%3A+escape', '\\href{page\%3A+with\%3A+escape}{inner}'),
+        ('page#target', '\\href{page\#target}{inner}')
+    ])
+    def test_link(self, target, output):
+        self._test_token('Link', output, target=target)
 
-    def test_autolink(self):
-        output = '\\url{target}'
-        self._test_token('AutoLink', output, target='target')
+    @parameterized.expand([
+        ('page', '\\url{page}'),
+        ('page%3A+with%3A+escape', '\\url{page\%3A+with\%3A+escape}'),
+        ('page#target', '\\url{page\#target}')
+    ])
+    def test_autolink(self, target, output):
+        self._test_token('AutoLink', output, target=target)
 
     def test_math(self):
         output = '$ 1 + 2 = 3 $'


pbodnar:
> If `\url{}` is used within macros like `\multicolumn` and if the link to
> be rendered contains `#` or `%` as typical confluence links like
> https://site.com/confluence/display/TEST/TEST-042%3A+Foo#target
> are bound to, pdflatex (from TeX Live 2019) fails with errors. This
> patch fixes that by escaping these two characters.

@choeppler, thank you for looking at this issue.

Regarding what characters to escape, when looking at this example LaTeX output:

```LaTeX
\href{https://site.com/confluence#something}{link}
```

... shouldn't we escape `\`, `{` and `}` as well (while replacing `\` for `\\` firstly)? I know it is not quite common to have these characters in a URL, yet it is practically possible.

pbodnar:
Regarding the commit message, it looks like you've got some redundant characters on the first line. It is also good to add issue number. I.e.:

```diff
-fix: escape special LaTeX characters in URLs``
+fix: escape special LaTeX characters in URLs (#114)
```

mistletoe/latex_renderer.py
```diff
@@ -57,15 +57,21 @@ def render_image(self, token):
...
+    def render_url(token):
+        return token.target.replace('%', '\\%') \
+                           .replace('#', '\\#')
+
     def render_link(self, token):
         self.packages['hyperref'] = []
         template = '\\href{{{target}}}{{{inner}}}'
         inner = self.render_inner(token)
-        return template.format(target=token.target, inner=inner)
+        return template.format(target=self.render_url(token),
```
pbodnar, line 69: OK, but what about escaping the `inner` part as well?

mistletoe/latex_renderer.py
```diff
@@ -57,15 +57,21 @@ def render_image(self, token):
         self.packages['graphicx'] = []
         return '\n\\includegraphics{{{}}}\n'.format(token.src)
 
+    @staticmethod
+    def render_url(token):
+        return token.target.replace('%', '\\%') \
+                           .replace('#', '\\#')
```
pbodnar, line 63: Can we rename it and unify this with already existing solution as seen at `HTMLRenderer.escape_url()`?

* it is a helper method, so it belongs to the end of the class
* the method should accept a string. This way, we can reuse it on another place.

test/test_latex_renderer.py
```diff
@@ -51,12 +51,20 @@ def test_image(self):
         self._test_token('Image', output, src='src')
 
     def test_link(self):
-        output = '\\href{target}{inner}'
-        self._test_token('Link', output, target='target')
+        for target, output in {
```
pbodnar, line 54: TODO for me: `output` should be `expected` - hopefully the last overlooked renaming within #185. I will do this for the whole file in a separate commit.

pbodnar, line 54: Dare to rewrite this to a parameterized test? :) Search for `@parameterized.expand` in the existing code - I've started to use that just recently...

pbodnar (changes requested):

pbodnar:
@choeppler, you still there? It would be great to have some feedback to my feedback. :) I think I can find the time to finish this change on my self otherwise...

choeppler:
@pbodnar , thank you for your review and sorry for the delay. I've been unexpectedly offline for some weeks but will try to address your review comments within the next few days. Ok?

pbodnar:
@choeppler, sure, I can wait. Good to hear you're back. :)

choeppler:
> Regarding what characters to escape, when looking at this example LaTeX output:
> 
> ```
> \href{https://site.com/confluence#something}{link}
> ```
> 
> ... shouldn't we escape `\`, `{` and `}` as well (while replacing `\` for `\\` firstly)? I know it is not quite common to have these characters in a URL, yet it is practically possible.

Well, looking at the first couple of answers to https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid, and also the RFCs these are based on, the three characters `\`, `{` and `}` should not occur in valid URLs. So I think we shouldn't escape those here. However, I just tested the complete valid character set with my LaTeX testcase and found '&' has to be escaped on the LaTeX-side, as well. I'll update the patch accordingly.

mistletoe/latex_renderer.py
choeppler: I just double-checked that the `inner` part is indeed properly escaped as is. `self.render_inner` eventually calls `self.render_raw_text` which takes care of that .

mistletoe/latex_renderer.py
choeppler: Done in 97bfce11

test/test_latex_renderer.py
choeppler: Done in 4231e80d

choeppler:
> Regarding the commit message, it looks like you've got some redundant characters on the first line. It is also good to add issue number. I.e.:
> 
> ```diff
> -fix: escape special LaTeX characters in URLs``
> +fix: escape special LaTeX characters in URLs (#114)
> ```

@pbodnar, I've added the changes addressing your review comments as additional commits to make the review process easier. Once we're done, I can prepare the merge by squashing and force-pushing this PR's commits. Plan to update the commit message as advised then.

pbodnar:
> > Regarding what characters to escape, when looking at this example LaTeX output:
> > 
> >     \href{https://site.com/confluence#something}{link}
> >   
> > ... shouldn't we escape `\`, `{` and `}` as well (while replacing `\` for `\\` firstly)? I know it is not quite common to have these characters in a URL, yet it is practically possible.
> 
> Well, looking at the first couple of answers to https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid, and also the RFCs these are based on, the three characters `\`, `{` and `}` should not occur in valid URLs. So I think we shouldn't escape those here. However, I just tested the complete valid character set with my LaTeX testcase and found '&' has to be escaped on the LaTeX-side, as well. I'll update the patch accordingly.

Well, yes, that's the _theory_ though. In practice, the characters `\`, `{` and `}` are still used sometimes (see e.g. [this answer](https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid/61174783#61174783) on the page you link to). An example accepted by Google: http://www.google.com/?q={a:1,b:2}.

Next, there is the correct finding of yours that all the possible characters with special meaning in LaTeX (as discussed within [#112](https://github.com/miyuchina/mistletoe/pull/112#issuecomment-921247077)) are already escaped within _render_raw_text()_ (which also includes escaping of the _label_ part of links):

```python
        return (token.content.replace('$', '\\$').replace('#', '\\#')
                             .replace('{', '\\{').replace('}', '\\}')
                             .replace('&', '\\&').replace('_', '\\_')
                             .replace('%', '\\%')
               ) if escape else token.content
```

So, to keep it simple, yet robust, and not to repeat ourselves in the code, what about simply using this existing method to also escape the URL part of the link? Or could it break something? I've tried to test links e.g. via https://latexeditor.lagrida.com/, but it doesn't seem to unescape e.g. `\&`, but I'm not sure if I can trust that tool. I couldn't find anything else working without registration fast.

It would be great if there was some LaTeX documentation that would direct us on this topic, but I can remember I couldn't find anything really useful some years ago. No that I don't use LaTeX myself...

choeppler:
> > Well, looking at the first couple of answers to https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid, and also the RFCs these are based on, the three characters `\`, `{` and `}` should not occur in valid URLs. So I think we shouldn't escape those here. However, I just tested the complete valid character set with my LaTeX testcase and found '&' has to be escaped on the LaTeX-side, as well. I'll update the patch accordingly.
> 
> Well, yes, that's the _theory_ though. In practice, the characters `\`, `{` and `}` are still used sometimes (see e.g. [this answer](https://stackoverflow.com/questions/1547899/which-characters-make-a-url-invalid/61174783#61174783) on the page you link to). An example accepted by Google: [http://www.google.com/?q={a:1,b:2}](http://www.google.com/?q=%7Ba:1,b:2%7D).

I see. (Funny: when copied from my browser's address field, the url is properly escaped: `http://www.google.com/?q=%7Ba:1,b:2%7D`...)

> Next, there is the correct finding of yours that all the possible characters with special meaning in LaTeX (as discussed within [#112](https://github.com/miyuchina/mistletoe/pull/112#issuecomment-921247077)) are already escaped within _render_raw_text()_ (which also includes escaping of the _label_ part of links):
> 
> ```python
>     def render_raw_text(self, token, escape=True):
>         return (token.content.replace('$', '\\$').replace('#', '\\#')
>                              .replace('{', '\\{').replace('}', '\\}')
>                              .replace('&', '\\&').replace('_', '\\_')
>                              .replace('%', '\\%')
>                ) if escape else token.content
> ```
> 
> So, to keep it simple, yet robust, and not to repeat ourselves in the code, what about simply using this existing method to also escape the URL part of the link? Or could it break something? I've tried to test links e.g. via https://latexeditor.lagrida.com/, but it doesn't seem to unescape e.g. `\&`, but I'm not sure if I can trust that tool. I couldn't find anything else working without registration fast.
> 
> It would be great if there was some LaTeX documentation that would direct us on this topic, but I can remember I couldn't find anything really useful some years ago. Note that I don't use LaTeX myself...

I haven't found any useful documentation either but I've been using pdflatex for a long time and just checked, how that implementation behaves if we escape all characters from the set escaped by `render_raw_text` in the URL part: 
- `\{` and `\}` if used in the URL part are not unescaped. If `{` and `}` are balanced, they can be used as is in the URL part, though.
- `\$` leads to a TEX stack overflow. `$` can be used in the URL part.
- rest is unescaped by LaTeX if escaped within the input URL part
- only '#' and '%' must be escaped in the URL part. Otherwise there are errors during LaTeX compilation if hyperref is used within multicolumn macros in a LaTeX table.

To sum up: the following hyperref-links are both displayed as and link to `http://a/A$#{}&_%` (NB: escaping the label part works as expected):
- `\href{http://a/A$\#{}\&\_\%}{http://a/A\$\#\{\}\&\_\%}` (max number of escapes in URL part that are properly dealt with by pdflatex)
- `\href{http://a/A$\#{}&_\%}{http://a/A\$\#\{\}\&\_\%}` (min number of necessary escapes in URL part for pdflatex to work without errors)

However, most of this is not so much of a problem if the URL-part is first escaped by `HTMLRenderer.escape_url` as proposed in [97bfce1](https://github.com/miyuchina/mistletoe/pull/190/commits/97bfce11b2572aea79b41f505a7cca76b1b57e8a), because then the only special characters that remain are `%`, `#`, `_`, and `&`, (the last two of which do not have to be escaped in the URL part but it doesn't seem to hurt either):
`<http://a/A$#{}&_%>` --> `\url{http://a/A\%24\#\%7B\%7D&amp;_\%}`

(My only concern with this solution is that if the links in the markdown-source are already url-escaped, then the `&` becomes double-escaped:
<http://a/A%24#%7B%7D&amp;_%> --> `\url{http://a/A\%24\#\%7B\%7D&amp;amp;_\%}`. But that could be noted down as a separate issue, I guess...)

test/test_latex_renderer.py
pbodnar: Great, thank you.

mistletoe/latex_renderer.py
pbodnar: OK.

mistletoe/latex_renderer.py
pbodnar: You're right, false alarm here.

pbodnar:
@choeppler, nice analysis, thanks for that. The code also looks +- OK.

Just this one:

> (My only concern with this solution is that if the links in the markdown-source are already url-escaped, then the `&` becomes double-escaped:
> `<http://a/A%24#%7B%7D&amp;_%>` --> `\url{http://a/A\%24\#\%7B\%7D&amp;amp;_\%}`. But that could be noted down as a separate issue, I guess...)

Yeah, that's the effect of the call to `html.escape()` inside `HTMLRenderer.escape_url()`. I would propose then not to call that function from LaTeX renderer code and simply duplicate only this part: `quote(raw, safe='/#:()*?=%@+,&;')`. I think we could make a shared function out of this expression (it is also used in Jira and XWiki renderers), but rather in a separate commit / MR.

pbodnar (approved):

pbodnar:
@choeppler, thank you!

test/test_latex_renderer.py
pbodnar: Done.

PR #192, opened 2023-08-17
refactor: use strict PascalCase for class names (#182)

### LINKED ISSUE #182: Use strict PascalCase for class names
The historically used "lenient" usage of the PascalCase (pascal case) convention in mistletoe is quite impractical when it comes to introducing more complex names. I suggest here to use the _strict_ variant which doesn't have this problem and leads to a more consistent code base.

E.g. change `HTMLRenderer` to `HtmlRenderer`, `HTMLBlock` to `HtmlBlock`.

For backwards compatibility, we can introduce aliases for the old names, marking them as deprecated.

pbodnar:
Fixes #182.

Giving this as a preview of the changes for a while, provided someone would like to give feedback.

Note that backwards compatibility is preserved - by introducing aliases for the old class names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,16 +122,16 @@ with open('dev-guide.md', 'r') as fin:
 ```
 
 Finally, here's how you would manually specify extra tokens via a renderer.
-In the following example, we use `HTMLRenderer` to render
-the AST. The renderer itself adds `HTMLBlock` and `HTMLSpan` tokens to the parsing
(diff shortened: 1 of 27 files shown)


pbodnar:
Note that names containing `XWiki` or `LaTeX` are left unchanged - I'm not sure if changing them to `Xwiki` and `Latex` is worth it, probably not?

PR #195, opened 2023-10-21
fix inline mathjax handling

chrisbresten:
previously it was converting mathjax unsupported inline math delimiters $ to block math delimiters $$ which are not equivalent. here it instead converts inline math delim $ to the mathjax supported inline math delim \\( \\) which have the expected equivalent behavior

### DIFF
--- a/mistletoe/contrib/mathjax.py
+++ b/mistletoe/contrib/mathjax.py
@@ -18,11 +18,11 @@ class MathJaxRenderer(HtmlRenderer, LaTeXRenderer):
 
     def render_math(self, token):
         """
-        Convert single dollar sign enclosed math environments to the \( \) syntax, to support
-        default mathjax settings which ignore single dollar signs for compatibility. 
-        https://docs.mathjax.org/en/latest/basic/mathematics.html#tex-and-latex-input
+        Convert single dollar sign enclosed math expressions to the ``\(...\)`` syntax, to support
+        the default MathJax settings which ignore single dollar signs as described at
+        https://docs.mathjax.org/en/latest/basic/mathematics.html#tex-and-latex-input.
         """
-        if token.content.startswith('$$'): 
+        if token.content.startswith('$$'):
             return self.render_raw_text(token)
         return '\\({}\\)'.format(self.render_raw_text(token).strip('$'))
 
--- a/mistletoe/contrib/mathjax.py
+++ b/mistletoe/contrib/mathjax.py
@@ -18,9 +18,12 @@ class MathJaxRenderer(HtmlRenderer, LaTeXRenderer):
 
     def render_math(self, token):
         """
-        Convert single dollar sign enclosed expressions to \(\) inline math delimiter for mathjax
+        Convert single dollar sign inline math expressions
+        (not supported by mathjax by default) to \( \) inline math delimiter enclosed 
+        expressions(supported by mathjax), let double dollar signs remain, to be 
+        rendered as blockmath by mathjax.
         """
-        if token.content.startswith('$$') or token.content.startswith('\\[') or token.content.startswith('\\('):
+        if token.content.startswith('$$'): 
             return self.render_raw_text(token)
         return '\({}\)'.format(self.render_raw_text(token).strip('$'))
 
--- a/mistletoe/contrib/mathjax.py
+++ b/mistletoe/contrib/mathjax.py
@@ -18,11 +18,11 @@ class MathJaxRenderer(HtmlRenderer, LaTeXRenderer):
 
     def render_math(self, token):
         """
-        Ensure Math tokens are all enclosed in two dollar signs.
+        Convert single dollar sign enclosed expressions to \(\) inline math delimiter for mathjax
         """
-        if token.content.startswith('$$'):
+        if token.content.startswith('$$') or token.content.startswith('\\[') or token.content.startswith('\\('):
             return self.render_raw_text(token)
-        return '${}$'.format(self.render_raw_text(token))
+        return '\({}\)'.format(self.render_raw_text(token).strip('$'))
 
     def render_document(self, token):
         """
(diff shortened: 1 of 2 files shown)


mistletoe/contrib/mathjax.py
```diff
@@ -18,11 +18,11 @@ def __init__(self, **kwargs):
 
     def render_math(self, token):
         """
-        Ensure Math tokens are all enclosed in two dollar signs.
+        Convert single dollar sign enclosed expressions to \(\) inline math delimiter for mathjax
         """
-        if token.content.startswith('$$'):
+        if token.content.startswith('$$') or token.content.startswith('\\[') or token.content.startswith('\\('):
```
pbodnar, line 23: Not sure about the extended condition here, because `Math` token should only be parsed from text enclosed in `$$` or `$` in mistletoe:

```python
class Math(span_token.SpanToken):
    pattern = re.compile(r'(\${1,2})([^$]+?)\1')
    parse_inner = False
    parse_group = 0
```

So the extended condition would make sense if we also extended the existing `pattern`. But do we want it?

pbodnar (changes requested):
Hi @chrisbresten, thank you for the PR. I've got some suggestions and questions.

Also, there is this unit test https://github.com/miyuchina/mistletoe/blob/v1.2.1/test/test_contrib/test_mathjax.py which needs amendments, so that it doesn't newly fail. (BTW its testing markdown is pretty "artificial", I think it would deserve quite a rewrite - at minimum to use some "real" math expressions like `2^1` in it.)

mistletoe/contrib/mathjax.py
```diff
@@ -18,11 +18,11 @@ def __init__(self, **kwargs):
 
     def render_math(self, token):
         """
-        Ensure Math tokens are all enclosed in two dollar signs.
+        Convert single dollar sign enclosed expressions to \(\) inline math delimiter for mathjax
```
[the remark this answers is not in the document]
chrisbresten: yes this is more thorough. The context here is that mathjax doesn't support single dollar signs because it causes too much fuss as it is too common in normal speech on web pages where matjax is supposed to live. but $ was always inline math and $$ block math from latex syntax, which later added in ``\(\)`` and ``\[ \]`` as alternatives that are easier to parse and read/debug because they are polarized. 

so converting the single dollar signs to the new syntax doesn't represent any change or imposition on the rendering, it is just adjusting standard syntax to be compatible with the modified syntax for mathjax specifically.

mistletoe/contrib/mathjax.py
chrisbresten: I didnt want to mess with the parsing, only the rendering, and erroneously assumed they were supported so i added them in there. it is a question of mistletoe and markdown as a standard if it should support ``\( \) \[ \]`` and if that will cause other problems like some collision in syntax(i am not sure). 

that being said, i think githubs markdown rendering supports them, and they are both mathjax and latex syntax, so tokenizing them as math with the properly modified regex is all it takes to make them render cleanly in both.

chrisbresten: \[ 
\sum\limits_{x=1}^{5}x^2
\]

inline \(\int x^3\)

mistletoe/contrib/mathjax.py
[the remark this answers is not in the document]
pbodnar: OK, thanks for the confirmation.

mistletoe/contrib/mathjax.py
pbodnar: OK, then I think we can just leave the original `if token.content.startswith('$$'):` condition in here.

pbodnar:
@chrisbresten, would you like to finish the small "cosmetic modifications" as suggested above, or can I take it over? :)

chrisbresten:
> @chrisbresten, would you like to finish the small "cosmetic modifications" as suggested above, or can I take it over? :)

hi, i am sorry for the slow response. I am not that well versed in the style conventions and cleanliness standards for production in this context so i had kind of deferred to you implicitly. I should have stated this.

chrisbresten:
i checked my local copy and found changes to unit tests to reflect the new behavior properly. I defer to you @pbodnar for any changes that help maintain standards and style consistent with the codebase at large

mistletoe/contrib/mathjax.py
```diff
@@ -18,11 +18,11 @@ def __init__(self, **kwargs):
...
         """
-        if token.content.startswith('$$'):
+        if token.content.startswith('$$') or token.content.startswith('\\[') or token.content.startswith('\\('):
             return self.render_raw_text(token)
-        return '${}$'.format(self.render_raw_text(token))
+        return '\({}\)'.format(self.render_raw_text(token).strip('$'))
```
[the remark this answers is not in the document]
pbodnar: Done.

mistletoe/contrib/mathjax.py
pbodnar: Done - we don't extend the parsed Math syntax for now.

pbodnar:
@chrisbresten, thanks once again. I did the final amendments and hopefully didn't break anything. ;)

pbodnar:
BTW I have found this cool "cheatsheet" for the MathJax syntax: https://math.meta.stackexchange.com/questions/5020/mathjax-basic-tutorial-and-quick-reference/.

chrisbresten:
> BTW I have found this cool "cheatsheet" for the MathJax syntax: https://math.meta.stackexchange.com/questions/5020/mathjax-basic-tutorial-and-quick-reference/.

thankyou. I write a lot of latex and there can be a lot of variation in what kind of syntax people choose to use out of habbit. there are a lot of options for some functionality due to improvements manifesting as new commands in order to leave the old ones for backward compatibility. also a lot of packages people use so ubiquitously that they forget the corresponding functionality is not builtin. mathjax seems to do a good job at keeping a pulse on this and providing a math environment that is compliant with this implicit standard

PR #202, opened 2023-12-02
feat(markdown-renderer): option to normalize spacing after the list marker

pbodnar:
This follows #197 and adds the possibility to switch ON the original renderer behavior when the space after list marker (called leader in mistletoe) collapsed just to 1 space.

### DIFF
--- a/mistletoe/markdown_renderer.py
+++ b/mistletoe/markdown_renderer.py
@@ -92,11 +92,22 @@ class MarkdownRenderer(BaseRenderer):
 
     _whitespace = re.compile(r"\s+")
 
-    def __init__(self, *extras, max_line_length: int = None):
+    def __init__(
+        self,
+        *extras,
+        max_line_length: int = None,
+        normalize_whitespace=False
+    ):
         """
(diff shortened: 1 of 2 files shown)


pbodnar:
@allets, @anderskaplan, this PR follows the discussions from #197, provided you would like to give feedback. If no one protests in a few days, I would merge this to the master. :)

anderskaplan:
lgtm!
just one thought about the naming of the flag, which is now very specfic. I believe that there are basically two modes you might want to run the Markdown renderer in: either you want it to stay as faithful to the original as possible, or you want the generated Markdown to be as regular (normalized) as possible. I can imagine that other options like this one will come, and maybe it's better to bundle them together with a single "normalize whitespace" option.

pbodnar:
@anderskaplan, thanks for your feedback. So I would rename the flag to `normalize_whitespace`, with an outlook that we can gradually take this flag into account in more places of the renderer, right? I haven't studied the renderer thoroughly... :P

allets:
@pbodnar, good!
Thanks for your commit that fixed my omission.
I like the idea of the generalization.
It works fine.

PR #203, opened 2023-12-09
feat: allow just 1 dash in Table delimiter cells (#131)

### LINKED ISSUE #131: Make tables GFM-compliant - support less than 3 dashes in delimiters
As per the specification for [GitHub flavored Markdown](https://github.github.com/gfm/#tables-extension-) (GFM) the following markdown

```
| foo |
|-|
| bar|
```

should be rendered as a table:

| foo |
|-|
| bar|

However mistletoe does not do so since it expects at least three dashes:

https://github.com/miyuchina/mistletoe/blob/8447a42ffa9b83cd55b05b206e36113592723248/mistletoe/block_token.py#L643

It would probably be a good idea to ensure that mistletoe passes all 8 test cases defined in the GFM table specification.

pbodnar:
Fixes #131.

### DIFF
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -676,7 +676,7 @@ class ListItem(BlockToken):
 
 class Table(BlockToken):
     """
-    Table token.
+    Table token. See its GFM definition at <https://github.github.com/gfm/#tables-extension->.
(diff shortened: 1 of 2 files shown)


pbodnar:
Leaving this open for community feedback for a few days.

PR #204, opened 2023-12-09
feat(html-renderer): option to skip HTML tokens parsing (#74)

### LINKED ISSUE #74: HTML output without inline HTML blocks

pbodnar:
This should resolve #74.

### DIFF
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -18,7 +18,14 @@ class HtmlRenderer(BaseRenderer):
 
     See mistletoe.base_renderer module for more info.
     """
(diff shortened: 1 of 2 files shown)


pbodnar:
Leaving this open for community feedback for a few days.

PR #206, opened 2023-12-17
feat: add parent attribute (#71)

### LINKED ISSUE #71: Get parent token

pbodnar:
This resolves #71 and is a continuation / replacement for PR #72.

### DIFF
--- a/dev-guide.md
+++ b/dev-guide.md
@@ -22,6 +22,10 @@ In this category you will find tokens like `RawText`, `Link`, and `Emphasis`.
 Block tokens may have block tokens, span tokens, or no tokens at all as children
 in the AST; this depends on the type of token. Span tokens may *only* have span
 tokens as children.
+Every token has properties called `children` and `parent` that can be used
+for traversing the nodes hierarchy.
+
+### Viewing the AST
 
 In order to see what exactly gets parsed, one can simply use the `AstRenderer`
 on a given markdown input, for example:
--- a/mistletoe/ast_renderer.py
+++ b/mistletoe/ast_renderer.py
@@ -42,7 +42,7 @@ def get_ast(token):
         node[attrname] = getattr(token, attrname)
     if 'header' in vars(token):
         node['header'] = get_ast(getattr(token, 'header'))
-    if 'children' in vars(token):
+    if token.children is not None:
         node['children'] = [get_ast(child) for child in token.children]
     return node
 
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -49,7 +49,7 @@ class HtmlRenderer(BaseRenderer):
         super().__exit__(*args)
 
     def render_to_plain(self, token) -> str:
-        if hasattr(token, 'children'):
+        if token.children is not None:
             inner = [self.render_to_plain(child) for child in token.children]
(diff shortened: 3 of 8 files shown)


pbodnar:
I'm hopefully choosing the simplest solution for introducing this feature. By making both the `children` and `parent` attribute a `@property`, we can extend the existing code and centralize the solution without much refactoring - although the API won't remain fully backwards compatible (`hasattr(...)` needs to be replaced by `... is not None` etc.).

An alternative approach would possibly be to pass the `parent` to every token constructor up to the `Token` class, but if I'm not mistaken, this would require much more changes - which would newly require calling `super().__init__(...)` in all the token classes. And I'm not sure it is a good time for doing that now.

Another thing is the **performance penalty** - running the benchmark (`python test/benchmark.py mistletoe` or running it via [mprof](https://pypi.org/project/memory-profiler/) to also see memory usage) shows me that mistletoe gets 4% slower and uses +13% memory when recording the children's `parent`. These are not great results, OTOH 1) it shouldn't be of such a big deal nowadays, 2) it cannot be probably implemented in a more efficient way (or can it?), 3) all the other competing tools (markdown, commonmark and its "de-facto replacement" markdown-it-py) seem to record token's parent (with no option to turn this off).

All in all, if someone (like @anderskaplan? ;)) doesn't propose a better solution, I would finish this PR (adding some docs and updating the benchmark results (yes, we should automate running those benchmarks...)) and merge it to the master branch. I think this would nicely complement the recent PR #188...

pbodnar:
I've decided to merge this into the master and update the benchmark results later on (there is more to update).

_Here are the COMPATIBILITY REMARKS:_

* As the `children` _attribute_ changed to _property_, existing code needs to be changed like this:
  ```diff
  -hasattr(token, 'children')
  +token.children is not None

  # ...
  
  -'children' in vars(token)
  +token.children is not None
  
  # ...
  
  -getattr(token, 'children', [])
  +token.children or []
  ```

vallentin:
Hey, sorry for not coming back to this. But since this has been merged, then I guess we can close my PR #72? :)

pbodnar:
@vallentin, you are right, thanks for the reminder. 👍

PR #215, opened 2024-04-22
fix(markdown-renderer): preserve list margin spacing

### LINKED ISSUE #213: Keep list item indentation in MarkdownRenderer
Hello!  I'm wondering if it's possible to adjust the `MarkdownRenderer`'s output spacing of indented list items to preserve four spaces of indentation in nested lists.

Using `mistletoe==1.3.0` on the following input file:
```markdown
1. A
    - B (4 leading spaces)
```
I obtain the following output from running `mistletoe test.md --renderer mistletoe.markdown_renderer.MarkdownRenderer`:

```markdown
1. A
   - B (3 leading spaces!)
```

It seems like the 3 space indentation level on the nested list is coming from the `prepend` attribute, which is set such that continuation lines are aligned with the content of the previous line.  This sort of makes sense, given that the inner list is really a child of the first list item.

However, I'd like the nested list item to render with the original 4 spaces of indentation, because I'm using mistletoe to preprocess some markdown for a downstream processor (mkdocs/python-markdown) which is fairly rigid in wanting 4 spaces or a tab for any block-level elements nested in a list.

I'm new to the codebase: poking around the `MarkdownRenderer` class and `render_list_item()`implementation, it isn't obvious how one could introduce an option like this.  It seems a bit tricky to recognize that when the child of a ListItem is another List, the prepend logic should preserve the indentation of the leader rather than treating it as a continuation line.

Of course one workaround is to carefully format your lists so that the prepend spacing ends up being 4:

```markdown
1.  A (content is padded out with an extra space)
    - B (so the indented leader ends up with 4 spaces!)
```

But I am working with a boatload of legacy markdown docs and would appreciate a solution that was a bit more robust to spacing between the leader and content.  Any ideas for quick monkeypatches or other workarounds would be greatly appreciated!  Thanks so much for the wonderful open source project!

nijel:
The extra spacing can be before of after leader, so try to preserve it on both positions.

Fixes #213

### DIFF
--- a/test/test_markdown_renderer.py
+++ b/test/test_markdown_renderer.py
@@ -138,11 +138,11 @@ class TestMarkdownRenderer(unittest.TestCase):
         ]
         output = self.roundtrip(input)
         expected = [
-            "22) *emphasized list item*\n",
-            "96) \n",
-            "128) here begins a nested list.\n",
-            "     + apples\n",
-            "     + bananas\n",
+            "  22) *emphasized list item*\n",
+            "  96) \n",
+            " 128) here begins a nested list.\n",
+            "       + apples\n",
+            "       + bananas\n",
         ]
         self.assertEqual(output, "".join(expected))
 
@@ -157,8 +157,7 @@ class TestMarkdownRenderer(unittest.TestCase):
         output = self.roundtrip(input)
         self.assertEqual(output, "".join(input))
 
-    # we don't currently support keeping margin indentation:
-    def test_list_item_margin_indentation_not_preserved(self):
+    def test_list_item_margin_indentation_preserved(self):
         # 0 to 4 spaces of indentation from the margin
         input = [
             "- 0 space: ok.\n",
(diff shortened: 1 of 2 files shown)


test/test_markdown_renderer.py
```diff
@@ -577,8 +576,8 @@ def test_wordwrap_paragraph_in_list(self):
                 "1. List item\n"
                 "2. A second list item\n"
                 "   including:\n"
-                "   * Nested list. This is\n"
-                "     a continuation line\n"
+                "   * Nested list. This is a\n"
+                "     continuation line\n"
```
nijel, line 580: I have no clue why this has changed.

pbodnar: Yeah, I struggled to find the cause this change as well for a while, then I found it in the calculation of `max_child_line_length` - see my fix in the code review commit.

pbodnar:
@nijel, thanks for your PR! I've found some small details and fixed them right away - can you please review my commit and squash it to your original commit if you find it OK?

nijel:
Looks good, I've just squashed that.

PR #224, opened 2024-08-28
fix(latex-renderer): thematic break on new line

choeppler:
Given the following markdown snippet
```md
line

---

other line
```
and creating a pdf from the rendered LaTeX with pdflatex, version pdfTeX 3.141592653-2.6-1.40.25 (TeX Live 2023/Debian)), I get
image

However, I'd expect the horizontal rule to be on a separate line as shown in the following image
image

The attached patch fixes that by adding an empty line before the `\hrulefill`.

To reproduce, run the following script and create a pdf from the resulting LaTeX-file via `pdflatex test.tex`:
```python
#!/bin/env python3

import mistletoe
from mistletoe.latex_renderer import LaTeXRenderer

def render(f):
   with open(f, 'r') as fp:
     return mistletoe.markdown(fp, LaTeXRenderer)

test = """
line

---

other line
"""

open('test.md', 'w').write(test)
open('test.tex', 'w').write(render('test.md'))
```

### DIFF
--- a/mistletoe/latex_renderer.py
+++ b/mistletoe/latex_renderer.py
@@ -158,7 +158,7 @@ class LaTeXRenderer(BaseRenderer):
 
     @staticmethod
     def render_thematic_break(token):
-        return '\\hrulefill\n'
+        return '\n\\hrulefill\n'
 
     @staticmethod
     def render_line_break(token):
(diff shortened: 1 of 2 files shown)


pbodnar:
Seems to make sense, thank you, @choeppler! 👍

PR #226, opened 2024-09-29
docs: fix typos

kianmeng:
Found via `codespell -H`


pbodnar:
@kianmeng, thank you.

I would omit changing the `docs/contributing.html` file though, as this file is generated (`python -m docs`) -> I'm thinking about removing the generated html files altogether, while also fixing the generator that currently doesn't output `\n` in between the top level `<body>` elements...

pbodnar:
Merged manually (omitting the html file): dd8de86fa.

In the previous commits, I have also resolved the historical problems within the `docs` folder as mentioned above...

kianmeng:
@pbodnar Noted, and sorry I missed out the message.

PR #236, opened 2025-02-04
Adding DefinitionListBlock

Lucas-C:

PR #238, opened 2025-02-18
Allowing BlockToken.start() to receive a second optional argument: lines, a FileReader

Lucas-C:
First PR to implement this feature request: https://github.com/miyuchina/mistletoe/pull/236

## Benchmark
On my computer, on the `master` branch:
```
$ python test/benchmark.py mistletoe
Test document: test/samples/syntax.md
Test iterations: 1000
Running tests with mistletoe...
===============================
mistletoe: 24.538773534994107
```

On my computer, on this PR branch:
```
$ python test/benchmark.py mistletoe
Test document: test/samples/syntax.md
Test iterations: 1000
Running tests with mistletoe...
===============================
mistletoe: 22.056132343015634
```

### DIFF
--- a/mistletoe/block_tokenizer.py
+++ b/mistletoe/block_tokenizer.py
@@ -99,7 +99,7 @@ def tokenize_block(iterable, token_types, start_line=1):
                 is_start = token_type.start(line, lines.reader())
             else:
                 raise NotImplementedError(
-                    f"start() method of {token_type} has an incorrect number of parameters: {func_args_count}"
+                    "start() method of {} has an incorrect number of parameters: {}".format(token_type, func_args_count)
                 )
             if is_start:
                 line_number = lines.line_number() + 1


pbodnar:
> ## Benchmark
> 
> On my computer, on the `master` branch:
> 
> ```
> $ python test/benchmark.py mistletoe
> Test document: test/samples/syntax.md
> Test iterations: 1000
> Running tests with mistletoe...
> ===============================
> mistletoe: 24.538773534994107
> ```
> 
> On my computer, on this PR branch:
> 
> ```
> $ python test/benchmark.py mistletoe
> Test document: test/samples/syntax.md
> Test iterations: 1000
> Running tests with mistletoe...
> ===============================
> mistletoe: 22.056132343015634
> ```

Can you please check again? Because it should be the other way round. On my computer, it is 14,5 vs 16,2, which is 5% performance penalty. That is relatively a lot, while the code you've added seems to be already optimized:

```python
            bound_func = token_type.start
            # Using .__code__.co_argcount has some shortcomings
            # (it ignores *args & parameters with default values),
            # but it is a lot faster than using inspect.signature():
            func_args_count = bound_func.__code__.co_argcount
            if isinstance(bound_func, MethodType):
                func_args_count -= 1
```

Further optimization strategies which come to my mind:

1. Cache the results of "args count findings" in a dictionary - are there some technical obstacles here and will it be really faster?
2. Make this whole new feature optional - this would make the config/API more complicated though.
3. Always pass all the lines to the start methods _somehow_ (efficiently) - but this would mean a change to the API of existing token classes, which of course we try to avoid by this PR.

Lucas-C:
> Can you please check again? Because it should be the other way round. On my computer, it is 14,5 vs 16,2, which is 5% performance penalty.

I just made another test:
```
$ git checkout master
Switched to branch 'master'
Your branch is up to date with 'origin/master'.
$ python test/benchmark.py mistletoe
Test document: test/samples/syntax.md
Test iterations: 1000
Running tests with mistletoe...
===============================
mistletoe: 14.430148589017335

$ git checkout tokenize_block-pass-lines-to-start
Switched to branch 'tokenize_block-pass-lines-to-start'
Your branch is up to date with 'origin/tokenize_block-pass-lines-to-start'.
$ python test/benchmark.py mistletoe
Test document: test/samples/syntax.md
Test iterations: 1000
Running tests with mistletoe...
===============================
mistletoe: 15.87390842998866

$ python --version
Python 3.10.12
```

So yeah, there may be a slight performance penalty (10% based on that test).

But I'm a bit dubious regarding this benchmark: isn't it very dependent on the hardware, and the current CPU usage, that can be busy due to other tasks? (_.e.g. running a web browser aside)

I ran it again, with the code from by branch, and this time got `14.960468987002969` as a result...
That's just a 1% difference.

Lucas-C:
> Further optimization strategies which come to my mind:
>  1. Cache the results of "args count findings" in a dictionary - are there some technical obstacles here and will it be really faster?

I had some caching mechanism in the last commit.
Based on `test/benchmark.py`, it does not make a significant improvement, so I don't known if it's really worth it.

>  2. Make this whole new feature optional - this would make the config/API more complicated though.

Do you have any suggestion in mind?
Would that mind introducing a configuration parameter for the parsing phase?

>  3. Always pass all the lines to the start methods _somehow_ (efficiently) - but this would mean a change to the API of existing token classes, which of course we try to avoid by this PR.

I think that's your call really, as a maintainer, not mine 🙂

pbodnar:
> I just made another test:
> 
> ...
> 
> So yeah, there may be a slight performance penalty (10% based on that test).
> 
> But I'm a bit dubious regarding this benchmark: isn't it very dependent on the hardware, and the current CPU usage, that can be busy due to other tasks? (_.e.g._ running a web browser aside)
> 
> I ran it again, with the code from by branch, and this time got `14.960468987002969` as a result... That's just a 1% difference.

Well, the tests are more consistent on my computer - still around 5% difference. To decrease the change of the results being random:

1) Run the same test at least twice, or thrice for sure.
2) Switch off all background activity, mainly if it doesn't use CPU in a constant way.

pbodnar:
> > Further optimization strategies which come to my mind:
> > 
> > 1. Cache the results of "args count findings" in a dictionary - are there some technical obstacles here and will it be really faster?
> 
> I added some caching mechanism in the last commit. Based on `test/benchmark.py`, it does not make a significant improvement, so I don't known if it's really worth it.

Yeah, confirmed, there doesn't seem to be major (if any) performance boost from the caching, so probably not worth it.

> > 2. Make this whole new feature optional - this would make the config/API more complicated though.
> 
> Do you have any suggestion in mind? Would that mean introducing a configuration parameter for the parsing phase?

Not yet. :P

> > 3. Always pass all the lines to the start methods _somehow_ (efficiently) - but this would mean a change to the API of existing token classes, which of course we try to avoid by this PR.
> 
> I think that's your call really, as a maintainer, not mine 🙂

True. I think we should probably choose this way in the end. So I would release the current version from the master branch and we would do a backwards-incompatible (yet quite easily "adaptable") one from this one. Hope that makes sense.

-----

> PS: `black` is mentioned in [CONTRIBUTING.md](https://github.com/miyuchina/mistletoe/blob/master/CONTRIBUTING.md) but the code is not actually formatted with `black`, so will not perform this formatting in this PR, as it would cause a lot of unrelated changes.

Agreed. The "unwritten rule" is not to reformat existing code, so that we avoid these "pseudo changes" to the code. (Luckily VS Code and the Black formatter have learnt to format just a selection recently.)

Lucas-C:
> > > 5\. Always pass all the lines to the start methods _somehow_ (efficiently) - but this would mean a change to the API of existing token classes, which of course we try to avoid by this PR.
> > 
> > 
> > I think that's your call really, as a maintainer, not mine 🙂
> 
> True. I think we should probably choose this way in the end. So I would release the current version from the master branch and we would do a backwards-incompatible (yet quite easily "adaptable") one from this one. Hope that makes sense.

Alright, makes sense 🙂

Do you need me to do antything more regarding this PR?

pbodnar:
@Lucas-C, thank you, I guess that's all for now regarding the original idea of having the `lines` argument _optional_. Do you think you would find the time to prepare and benchmark-test the "API-breaking" variant? I expect that if an instance of the newly introduced `FileReader` is being created and passed every time, that would also incur some overhead, but maybe not that big...?

mistletoe/block_tokenizer.py
```diff
@@ -78,7 +89,19 @@ def tokenize_block(iterable, token_types, start_line=1):
...
+            # * a start() method accepting only a single `line` parameter
+            # * a start() method accepting both a `line` and a `lines` parameter, e.g. DefinitionList
+            func_args_count = token_type.start_args_count()
+            if func_args_count == 1:
+                is_start = token_type.start(line)
+            elif func_args_count == 2:
+                is_start = token_type.start(line, lines.reader())
+            else:
+                raise NotImplementedError(
+                    f"start() method of {token_type} has an incorrect number of parameters: {func_args_count}"
```
pbodnar, line 102: Note: While we still support Python 3.5 (pretty old, right? :)), we need to rewrite this to the old style:

```suggestion
                    "start() method of {} has an incorrect number of parameters: {}".format(token_type, func_args_count)
```

Lucas-C:
> @Lucas-C, thank you, I guess that's all for now regarding the original idea of having the `lines` argument _optional_. Do you think you would find the time to prepare and benchmark-test the "API-breaking" variant? I expect that if an instance of the newly introduced `FileReader` is being created and passed every time, that would also incur some overhead, but maybe not that big...?

I opened PR to that effect: https://github.com/miyuchina/mistletoe/pull/241
The description includes benchmark results from my computer.

Also, I thought about another option to go forward:
switch from having a single `line` argument passed to `Token.start()` to having a single `lines` argument,
that would be an instance of a new `LineReader` class,
that inherits from `str` to keep backward compatibility.

What do you think about this?

pbodnar:
> > @Lucas-C, thank you, I guess that's all for now regarding the original idea of having the `lines` argument _optional_. Do you think you would find the time to prepare and benchmark-test the "API-breaking" variant? I expect that if an instance of the newly introduced `FileReader` is being created and passed every time, that would also incur some overhead, but maybe not that big...?
> 
> I opened PR to that effect: #241 The description includes benchmark results from my computer.

Thank you!

> Also, I thought about another option to go forward: switch from having a single `line` argument passed to `Token.start()` to having a single `lines` argument, that would be an instance of a new `LineReader` class, that inherits from `str` to keep backward compatibility.
> 
> This way we could have all end-users subclasses of `mistletoe.Token` still work properly when they upgrade to the new release of `mistletoe`.
> 
> What do you think about this?

Interesting idea, but not sure if this kinda-hack wouldn't "bite" us later on somewhere. I guess there would be also some performance impact...

PR #241, opened 2025-03-10
Allowing BlockToken.start() to receive a second optional argument: lines, a FileReader

Lucas-C:
This PR is an alternative to PR #238, that @pbodnar suggested we test in comment https://github.com/miyuchina/mistletoe/pull/238#issuecomment-2708740175

The goal is to implement this feature request: https://github.com/miyuchina/mistletoe/pull/236

Changes introduced by this PR:
* `FileWrapper` now inherits from a new `FileReader` class
* new method `FileWrapper.reader()` that returns a FileReader with read-only access to the same file, positioned at the current index.
* `BlockToken.start()` now receives as a second optional argument a `lines` parameter, that is `FileReader` produced by calling `FileWrapper.reader()`

---

Benchmark results on my computer:

* on the `master` branch:
```
$ python test/benchmark.py mistletoe
Test document: test/samples/syntax.md
Test iterations: 1000
Running tests with mistletoe...
===============================
mistletoe: 19.028154349070974
```

* on this `tokenize_block-always-pass-lines-to-start` branch:
```
$ python test/benchmark.py mistletoe
Test document: test/samples/syntax.md
Test iterations: 1000
Running tests with mistletoe...
===============================
mistletoe: 20.68729400797747
```

To improve performances, we could get rid of the new `FileReader` class,
and simply pass the current `FileWrapper` instance as 2nd argument to `Token.start()`.

### DIFF
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -165,7 +165,7 @@ class Heading(BlockToken):
         super().__init__(content, span_token.tokenize_inner)
 
     @classmethod
-    def start(cls, line):
+    def start(cls, line, lines):
         match_obj = cls.pattern.match(line)
         if match_obj is None:
             return False
@@ -178,7 +178,7 @@ class Heading(BlockToken):
 
     @classmethod
     def check_interrupts_paragraph(cls, lines):
-        return cls.start(lines.peek())
+        return cls.start(lines.peek(), lines)
 
     @classmethod
     def read(cls, lines):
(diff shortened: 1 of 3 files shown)


pbodnar:
> To improve performances, we could get rid of the new `FileReader` class, and simply pass the current `FileWrapper` instance as 2nd argument to `Token.start()`.

Yeah, I think in the end, this will probably be the way to go. While also warning the implementers about the need to reset the wrapper to the original state after usage (using the combination of firstly calling get_pos() and later set_pos() on the wrapper). I think we could also extend the existing Dev Guide (https://github.com/miyuchina/mistletoe/blob/master/dev-guide.md#a-new-token) in this regard...

PR #270, opened 2026-04-15
jira_render - handle pipe in table cell

ritzk:


test/test_contrib/test_jira_renderer.py
```diff
@@ -215,6 +215,22 @@ def test_table_empty_cell(self):
 ||A||B||C||
 |1| |3|
 
+"""
+        self.markdownResultTest(markdown, expected)
+
+    def test_table_pipe_in_cell(self):
+        """
+        Empty cells need to have a space in them, see <https://jira.atlassian.com/browse/JRASERVER-70048>.
+        """
```
pbodnar: A copy-paste error.

test/test_contrib/test_jira_renderer.py
```diff
@@ -215,6 +215,22 @@ def test_table_empty_cell(self):
...
+        markdown = """\
+| A    | B |
+|-----------|
+| 1\\|1 | 2  |
```
pbodnar: Good one, but it would look better if the columns were aligned as in other test methods?

pbodnar (changes requested):
@ritzk, thank you, this looks basically good, just a few details in the tests to fix.

Also note that it is generally better to create a dedicated branch in your fork repo for the PR, e.g. `escape-pipes-in-jira-render`, but probably never mind in this case...

test/test_contrib/test_jira_renderer.py
ritzk: resolved

test/test_contrib/test_jira_renderer.py
ritzk: resolved

PR #271, opened 2026-04-27
fix: avoid crash on overlapping delimiters

nijel:
The fuzzing discovered crash with IndexError: string index out of range

### DIFF
--- a/mistletoe/core_tokens.py
+++ b/mistletoe/core_tokens.py
@@ -439,9 +439,10 @@ class Delimiter:
             self.number = self.end - self.start
             self.type = self.type[n:]
             return True
+        # Trim closing markers from the right so type matches start/end.
         self.end = self.end - n
         self.number = self.end - self.start
-        self.type = self.type[:n]
+        self.type = self.type[:-n]
         return True
 
     def closed_by(self, other):
--- a/test/test_core_tokens.py
+++ b/test/test_core_tokens.py
@@ -30,6 +30,7 @@ class TestCoreTokens(TestCase):
         self.assertEqual(delimiter.number, 1)
         self.assertEqual(delimiter.start, 5)
         self.assertEqual(delimiter.end, 6)
+        self.assertEqual(delimiter.type, '*')
 
     def test_delimiter_remove_right(self):
         delimiter = Delimiter(4, 6, 'abcd**')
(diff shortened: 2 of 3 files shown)


test/test_block_token.py
```diff
@@ -596,6 +596,16 @@ def test_auto_splitlines(self):
...
+    def test_issue_261_table_cell_asterisk_sequences_do_not_crash(self):
+        lines = [
+            '| Qiup | Xmfakdshhgvdj |\n',
+            '| --- | --- |\n',
+            '| **NMH9920F** | **IKXZXPTB****AJFHV *Pbjovp*VSN* |\n',
+        ]
+        document = block_token.Document(lines)
+        self.assertEqual(len(document.children), 1)
+        self.assertIsInstance(document.children[0], block_token.Table)
+
```
pbodnar: I think this test for is not necessary, because as it turned out, the problem is not specific (just) to tables and one of the problematic samples `**IKXZXPTB****AJFHV *Pbjovp*VSN*` is being tested within `test_issue_261_asterisk_sequences_do_not_crash()` already.

pbodnar (changes requested):
@nijel, thanks a lot for fixing this long standing bug. Once again, I will have one or more remarks to the tests though... ;) (just one for now from a quick code review)

pbodnar:
> The fuzzing discovered crash with IndexError: string index out of range

Good job. 👍  But, do you think you can also share why the bug got fixed by the change made:

```diff
-self.type = self.type[:n]
+self.type = self.type[:-n]
```

Also, what does `type` mean in the given context? The original code is not easy to read and I have never found the time to investigate fully how it works...

Ideally, you can put the explanation into the commit message body, if not also optionally into some inline comment(s)...

nijel:
The `type` stores all parsed delimiters and this is supposed to remove one removed delimiter, but it did remove all but one delimiter so it only worked for one nesting level. For example, removing 1 marker from the right side of `****` should leave `***`, but the old code left `*` while `number`, `start`, and `end` said 3 markers remained. Later parsing could then reduce that inconsistent `type` to "", and `closer.type[0]` crashed with `IndexError`.

I've updated the patch with a short comment as well.

nijel:
I've removed the extra text (somehow I missed that part of feedback before).

pbodnar:
> I've removed the extra text (somehow I missed that part of feedback before).

Thank you. Yet, the test still seems to be present? 🤷‍♂️

nijel:
Sorry, I forgot to commit before pushing :facepalm:

test/test_block_token.py
pbodnar: Test removed, thanks.

PR #274, opened 2026-04-27
fix: escape pipes inside table

### LINKED ISSUE #244: Markdown renderer doesn't preserve escaped | in the table.
Let's have a following Markdown file:

```markdown
| Left column | Right column                   |
|-------------|--------------------------------|
| left baz    | right foo \| between pipes \|  |
```

The escaping is correctly handled when rendering to HTML using `mistletoe test.md`:

```html
<table>
<thead>
<tr>
<th align="left">Left column</th>
<th align="left">Right column</th>
</tr>
</thead>
<tbody>
<tr>
<td align="left">left baz</td>
<td align="left">right foo | between pipes |</td>
</tr>
</tbody>
</table>
```

When using the Markdown renderer as `mistletoe -r mistletoe.markdown_renderer.MarkdownRenderer test.md`, the escaping is gone:

```markdown
| Left column | Right column                |
| ----------- | --------------------------- |
| left baz    | right foo | between pipes | |
```

nijel:
Ensure that pipes in the Markdow rendered tables is escaped.

Fixes #244

### DIFF
--- a/test/test_markdown_renderer.py
+++ b/test/test_markdown_renderer.py
@@ -338,6 +338,15 @@ class TestMarkdownRenderer(unittest.TestCase):
         output = self.roundtrip(input)
         self.assertEqual(output, "".join(input))
 
+    def test_table_escapes_pipes_in_cells(self):
+        input = [
+            "| Left column | Right column                  |\n",
+            "| ----------- | ----------------------------- |\n",
+            "| left baz    | right foo \\| between pipes \\| |\n",
+        ]
+        output = self.roundtrip(input)
+        self.assertEqual(output, "".join(input))
+
     def test_table_with_varying_column_counts(self):
         input = [
             "   |   header | x |  \n",
--- a/mistletoe/markdown_renderer.py
+++ b/mistletoe/markdown_renderer.py
@@ -512,7 +512,19 @@ class MarkdownRenderer(BaseRenderer):
         """
         Renders each table cell on a table row to text. No word wrapping.
         """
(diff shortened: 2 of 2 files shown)


test/test_markdown_renderer.py
```diff
@@ -338,6 +338,18 @@ def test_table(self):
...
+        input = [
+            "| Left column | Right column                   |\n",
+            "|-------------|--------------------------------|\n",
+            "| left baz    | right foo \\| between pipes \\|  |\n",
+        ]
+        output = self.roundtrip(input)
+        self.assertIn("right foo \\| between pipes \\|", output)
+        table = Document(output).children[0]
+        self.assertEqual(len(table.header.children), 2)
+        self.assertEqual(len(table.children[0].children), 2)
```
pbodnar: OK, but what about doing it the same as in `test_table` above, for consistency:

```python
        input = [
            "| Left column | Right column                  |\n",
            "| ----------- | ----------------------------- |\n",
            "| left baz    | right foo \\| between pipes \\| |\n",
        ]
        output = self.roundtrip(input)
        self.assertEqual(output, "".join(input))
```

I.e. I don't also think testing the _parser_ on the last 3 selected lines is really necessary in _this_ test.

pbodnar (changes requested):
@nijel, thanks a lot, just some details in the tests...

test/test_markdown_renderer.py
nijel: Good point, I've adjusted the test.

PR #275, opened 2026-05-05
fix: preserve line break before/after HTML span during word-wrap

razzeee:
When a closing HTML tag (e.g. </template>) appears on its own line in
the source, mistletoe parses the surrounding newlines as soft LineBreaks.
During word-wrap (max_line_length set), soft breaks are treated as
spaces and collapsed, causing the tag to be moved onto the end of the
previous line or the following content to be merged onto the tag's line.

We're seeing this break our html via weblate.
For e.g. in https://codeberg.org/DI-Day/website/pulls/139/files

### DIFF
--- a/test/test_markdown_renderer.py
+++ b/test/test_markdown_renderer.py
@@ -638,6 +638,28 @@ class TestMarkdownFormatting(unittest.TestCase):
                 "> > hills of hay\n"
             )
 
+    def test_wordwrap_paragraph_with_html_span_on_own_line(self):
+        with MarkdownRenderer() as renderer:
+            paragraph = block_token.Paragraph(
+                [
+                    "Some long sentence that will be reflowed by the word wrapper.\n",
+                    "</template>\n",
+                    "More text after the tag.\n",
+                ]
+            )
+
+            renderer.max_line_length = 50
+            lines = renderer.render(paragraph)
+
+            # then the closing tag must remain on its own line — not collapsed
+            # onto the end of the previous line, and not merged with the next line
+            assert lines == (
+                "Some long sentence that will be reflowed by the\n"
+                "word wrapper.\n"
+                "</template>\n"
+                "More text after the tag.\n"
+            )
+
     def test_wordwrap_tables(self):
         with MarkdownRenderer(max_line_length=30) as renderer:
             # given a markdown table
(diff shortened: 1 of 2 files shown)


test/test_markdown_renderer.py
```diff
@@ -653,3 +653,24 @@ def test_wordwrap_tables(self):
...
+            renderer.max_line_length = 80
+            lines = renderer.render(paragraph)
+
+            # then the closing tag must remain on its own line — not collapsed
+            # onto the end of the previous line, and not merged with the next line
+            assert lines == (
+                "Some long sentence that will be reflowed by the word wrapper.\n"
+                "</template>\n"
+                "More text after the tag.\n"
+            )
```
pbodnar, line 661: OK, but:

1) I think this test logically belongs after `test_wordwrap_paragraph_in_block_quote()`, not after `test_wordwrap_tables()`.

2) If we say we are testing wrapping, then we should also activate and test it for the first line:

```suggestion
    def test_wordwrap_paragraph_with_html_span_on_own_line(self):
        with MarkdownRenderer() as renderer:
            paragraph = block_token.Paragraph(
                [
                    "Some long sentence that will be reflowed by the word wrapper.\n",
                    "</template>\n",
                    "More text after the tag.\n",
                ]
            )

            renderer.max_line_length = 50
            lines = renderer.render(paragraph)

            # then the closing tag must remain on its own line — not collapsed
            # onto the end of the previous line, and not merged with the next line
            assert lines == (
                "Some long sentence that will be reflowed by the\n"
                "word wrapper.\n"
                "</template>\n"
                "More text after the tag.\n"
            )
```

pbodnar (changes requested):
@razzeee, thank you, LGTM. Having just some notes to the tests...

pbodnar:
> We're seeing this break our html via weblate.
> For e.g. in https://codeberg.org/DI-Day/website/pulls/139/files

Just a detail, but I can't see the described breakage there. Maybe the PR has changed since this report?

razzeee:
Yes, I've stripped the problematic tag from our website, as it was keeping the block from rendering in the frontend.

razzeee:
Changed accordingly

test/test_markdown_renderer.py
pbodnar: Addressed, thanks.

PR #276, opened 2026-05-14
fix(gihub_wiki): reduce backtracking in wiki links

### LINKED ISSUE #249: [Bugs] Protential ReDoS bug in `github_wiki.py`
### Description

Hi team, thanks for your great work! I think I found a bug that might lead to DDoS in the system
At line 14 in `sqlite_ext.py`
the regex ` re.compile(r"\[\[ *(.+?) *\| *(.+?) *\]\]")` is vulnerable to ReDoS when it is used ito parse user input

### PoC

I have a test file to stimulate the user input handled by this regex
```python3
import re
import time


if __name__ == "__main__":
    pattern = re.compile(r"\[\[ *(.+?) *\| *(.+?) *\]\]")
    for i in range(100,300,50):
        user_input = "[[" + " " * i  + "|" + " " * i + "]"
        start_time = time.time()
        _ = pattern.match(user_input)
        end_time = time.time()
        duration = end_time - start_time
        print(f"rep_count:{i} / duration:{duration:.4f} s")
        

```
The result is like
```
rep_count:100 / duration:2.4525 s
rep_count:150 / duration:35.6338 s
rep_count:200 / duration:77.9442 s
rep_count:250 / duration:199.0532s
...
```
The result system will hang for around 4 mins here with around 200 chars in the attack query.

If using mistletoe in a server setup, this bug may lead to high CPU usage or DoS risks if users submit malicious or resource-intensive repositories.

**Expected behavior**
The issue here is that both ` *` and `(.+?)` will eagerly try to match the space chars, and lead to a large amount of backtracing.

Maybe we can add a timeout or set a limit there?
For example, I've tried to modify the regex a little bit to
```
re.compile(r"\[\[ {0,20}(.+?) {0,20}\| {0,20}(.+?) {0,20}\]\]")
```
And the performance seems to be pretty good~

```python
rep_count:100 / duration:0.0409 s
rep_count:150 / duration:0.0743 s
rep_count:200 / duration:0.0968 s
rep_count:250 / duration:0.1183 s
```

nijel:
The lookahead avoids backtracking in invalid patterns while keeping full compatibility.

Fixes #249

### DIFF
--- a/mistletoe/contrib/github_wiki.py
+++ b/mistletoe/contrib/github_wiki.py
@@ -11,7 +11,8 @@ __all__ = ['GithubWiki', 'GithubWikiRenderer']
 
 
 class GithubWiki(SpanToken):
-    pattern = re.compile(r"\[\[ *(.+?) *\| *(.+?) *\]\]")
+    # In order to avoid inefficient backtracking, we use positive lookahead here.
+    pattern = re.compile(r"\[\[(?=[^\n|]*\|[^\n]*\]\]) *([^\n|]+?) *\| *([^\n]+?) *\]\]")
 
     def __init__(self, match):
         self.target = match.group(2)
--- a/test/test_contrib/test_github_wiki.py
+++ b/test/test_contrib/test_github_wiki.py
@@ -1,4 +1,6 @@
+import time
 from unittest import TestCase, mock
+from parameterized import parameterized
 from mistletoe import span_token, Document, token
 from mistletoe.span_token import tokenize_inner
 from mistletoe.contrib.github_wiki import GithubWiki, GithubWikiRenderer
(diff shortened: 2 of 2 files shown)


test/test_contrib/test_github_wiki.py
```diff
@@ -28,3 +29,29 @@ def test_render(self):
...
+                token = next(iter(tokenize_inner(source)))
+                self.assertIsInstance(token, GithubWiki)
+                self.assertEqual(token.target, target)
+                self.assertEqual(self.renderer.render_inner(token), label)
+
+    def test_malformed_input_is_rejected_quickly(self):
+        malformed_inputs = [
+            '[[' + ' ' * 10000 + '|' + ' ' * 10000 + ']',
+            '[[' + ' ' * 10000 + ']]',
+        ]
```
pbodnar: "Devil is hidden in the details": I think we don't want this to run like forever when an non-optimized regex is used. I tested on my machine that _without the lookahead_, for 100x it runs for 1.6s, while for 200x for 40s - for the first input. For the second one it is much faster, yet 10000x seems to be too high again: 6000x seems to be enough - runs for 30s.

```suggestion
        malformed_inputs = [
            '[[' + ' ' * 200 + '|' + ' ' * 200 + ']',
            '[[' + ' ' * 6000 + ']]',
        ]
```

Also note that for parameterized tests like this, we prefer using _parameterized_ in the project.

mistletoe/contrib/github_wiki.py
```diff
@@ -11,7 +11,8 @@
 
 
 class GithubWiki(SpanToken):
-    pattern = re.compile(r"\[\[ *(.+?) *\| *(.+?) *\]\]")
+    # The lookahead requires a pipe and closing marker to avoid capture backtracking.
+    pattern = re.compile(r"\[\[(?=[^\n|]*\|[^\n]*\]\]) *([^|\n]+?) *\| *([^\n]+?) *\]\]")
```
pbodnar: I guess this should be technically equivalent, while keeping the repeated char-sets consistent within the regex:

```suggestion
    pattern = re.compile(r"\[\[(?=[^\n|]*\|[^\n]*\]\]) *([^\n|]+?) *\| *([^\n]+?) *\]\]")
```

mistletoe/contrib/github_wiki.py
```diff
@@ -11,7 +11,8 @@
 
 
 class GithubWiki(SpanToken):
-    pattern = re.compile(r"\[\[ *(.+?) *\| *(.+?) *\]\]")
+    # The lookahead requires a pipe and closing marker to avoid capture backtracking.
```
pbodnar: OK, but I would rather focus on "why", like this (not sure if the word "capture" is crucial here, feel free to include it if it is):

```suggestion
    # In order to avoid inefficient backtracking, we use positive lookahead here.
```

pbodnar (changes requested):
@nijel, thanks a lot, just a few details again. :P

pbodnar:
@nijel, all OK now, thank you. 👍

PR #277, opened 2026-05-14
feat: modernize packaging

### LINKED ISSUE #255: Modernize project setup (add pyproject.toml, etc.)
See https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#modernize-setup-py-project.

This should include getting rid of using MANIFEST.in for manually specifying source distribution (sdist) includes and excludes (see e.g. #253). This can be achieved by using [setuptools-scm](https://pypi.org/project/setuptools-scm), which simply includes only versioned files in the sdist archive.

After migration to PEP-517/518, we can no longer import `mistletoe` (`mistletoe.__version__`) in setup.py. So we need to read mistletoe version in another way. While at it, we can use [setuptools-scm](https://pypi.org/project/setuptools-scm) for automatic version file generation.

nijel:
- use pyproject.toml to define the packaging
- use proper optional dependencies for extrra
- use dependency groups for devel deps
- use uv for the development workflow
- drop support for Python < 3.10

Fixes #255

Things I decided to leave out, as both are quite independent of this change:

- migration to setuptools-scm
- linting improvemnts

Things to consider:

- I've committed `uv.lock` so that CI is reproducible, but this should come up with automated dependency updates such as renovate or dependabot


pbodnar:
@nijel, thanks a lot, this looks as a promising first step. I will check it out - I need to make sure everything works as expected, get to know the `uv` tool etc. So it will take some time, but chatbots will hopefully help me in getting on board quickly... :)

pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+    {name = "Mi Yu", email = "hello@afteryu.me"},
+]
+keywords = ["markdown", "lexer", "parser", "development"]
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "License :: OSI Approved :: MIT License",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.5",
+    "Programming Language :: Python :: 3.6",
```
choeppler: Since these two versions are not tested, shouldn't they be left out?

```suggestion
```

nijel: I kept these as they are currently present so that the change is limited to the packaging workflow and not changing the metadata. Such a change should go together with `requires-python` change so that these two match.

nijel: PS: I think it's time to bump the minimal supported Python version to 3.10 as that is the oldest upstream supported version now (see https://www.python.org/downloads/), but I wanted to do that in a separate pull request.

.github/workflows/python-package.yml
```diff
@@ -15,47 +15,43 @@ on:
...
     - name: Set up Python ${{ matrix.python-version }}
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
-      env:
-        # necessary for older Python versions:
-        # (see https://github.com/actions/setup-python/issues/866)
-        PIP_TRUSTED_HOST: "pypi.python.org pypi.org files.pythonhosted.org"
+    - name: Set up uv
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
```
mgaitan: setup-uv uses inmutable release, so it's safe and recommended to pin to the current minor

```suggestion
      uses: astral-sh/setup-uv@v8.2.0
```

ref https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases

.github/workflows/python-package.yml
```diff
@@ -65,30 +61,28 @@ jobs:
...
     - name: Set up Python ${{ env.python_version }}
       uses: actions/setup-python@v5
       with:
         python-version: ${{ env.python_version }}
-      env:
-        # necessary for older Python versions:
-        # (see https://github.com/actions/setup-python/issues/866)
-        PIP_TRUSTED_HOST: "pypi.python.org pypi.org files.pythonhosted.org"
+    - name: Set up uv
+      uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
```
mgaitan: no need to use bot setup-python and setup-uv  . uv can discover or install python on its own. 

consider my package template as an example
https://github.com/mgaitan/python-package-copier-template/blob/main/project/.github/workflows/ci.yml.jinja#L38

cutting-a-release.md
```diff
@@ -5,13 +5,11 @@ For maintainers, here are the basic steps when creating a new release of mistlet
...
     * official documentation: [Packaging Python Projects](https://packaging.python.org/en/latest/tutorials/packaging-projects/)
-    * install / upgrade the build tool: `$ python -m pip install --upgrade build`
     * make sure there are no old relicts in the local "dist" folder
-    * build distribution archives (sdist and Wheel files in "dist/*"): `$ python -m build`
+    * build distribution archives (sdist and Wheel files in "dist/*"): `$ uv build`
     * upload the distribution archives to PyPi:
-        * install / upgrade Twine: `$ python -m pip install --upgrade twine`
         * if unsure, upload to test PyPi and/or test locally - see [docs](https://packaging.python.org/en/latest/tutorials/packaging-projects/#uploading-the-distribution-archives)
-        * do the upload to PyPi: `$ python -m twine upload dist/*`
+        * do the upload to PyPi: `$ uvx twine upload dist/*`
```
mgaitan: uv replaces twine via `uv publish` . 

https://docs.astral.sh/uv/guides/package/#publishing-your-package


Moreover if it is done in github action, the uploadd requires no secrets if the "trusted publishing" is setup https://docs.pypi.org/trusted-publishers/adding-a-publisher/ 

so just `uv build`  / `uv publish`  is enough
https://github.com/mgaitan/python-package-copier-template/blob/main/project/.github/workflows/cd.yml.jinja#L40-L41

CONTRIBUTING.md
```diff
@@ -113,3 +127,4 @@ I tweet [@mi_before_yu][twitter]. Also yell at me over [email][email].
 [conv-commits]: https://www.conventionalcommits.org/
 [black-formatter]: https://black.readthedocs.io/
 [flake8]: https://flake8.pycqa.org/
```
mgaitan, line 129: If the goal is modernize the package, I strongly recommend to replace black and flake8 by ruff

pyproject.toml
```diff
@@ -0,0 +1,79 @@
...
+requires = ["setuptools>=64"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "mistletoe"
+dynamic = ["version"]
+description = "A fast, extensible Markdown parser in pure Python."
+readme = "README.md"
+requires-python = ">=3.5,<4"
+license = {text = "MIT"}
```
mgaitan: This is the standard form 

```suggestion
license = "MIT"
license-files = ["LICENSE"]

```

ref https://peps.python.org/pep-0639/

.github/workflows/python-package.yml
nijel:
```suggestion
      uses: astral-sh/setup-uv@v8.2.0
```

.github/workflows/python-package.yml
```diff
@@ -65,30 +61,28 @@ jobs:
...
+      with:
+        version: "0.11.14"
```
nijel:
```suggestion
```

CONTRIBUTING.md
nijel: Yes, that's a reasonable move, but I didn't consider this in scope here. I wanted to keep this focused on packaging itself only. There are definitely more things to cleanup or modernize here (see previous discussion on supported Python versions).

cutting-a-release.md
nijel:
```suggestion
        * do the upload to PyPI: `$ uv publish dist/*`
```

nijel: I forgot to update it here, done now.

.github/workflows/python-package.yml
nijel: AFAIR setup-python primarily uses Python versions available on the runner, while uv downloads them, or has that changed?

pyproject.toml
nijel: Hmm, now I know why I didn't use it before: I doesn't work with ancient Python versions. But let's include Python version bump as well.

nijel:
In the end I've decided to include Python version bump as well to be able to use modern license specification in pyproject.toml.

mgaitan (approved):

.github/workflows/python-package.yml
mgaitan: by default uv will also use the available versions.

PR #280, opened 2026-06-17
fix: make remove_token idempotent so renderers can be re-instantiated (#262)

### LINKED ISSUE #262: Bug: mistletoe 1.5.0 - block_token.remove_token(block_token.Footnote) fails
I have a script which manipulates multiple MD files into a single file and modifies the content through the AST as it goes to get one single document with links to the internal chapters.

At some point I make a new instance of the renderer:
`mdrenderer = MarkdownRenderer()`

This then fails with:
```
  File "/Users/cyberwizzard/markdown/md_combiner.py", line 787, in main
    run(argv, logger)
    ~~~^^^^^^^^^^^^^^
  File "/Users/cyberwizzard/markdown/md_combiner.py", line 714, in run
    mdrenderer = MarkdownRenderer()
  File "/Users/cyberwizzard/.cache/uv/environments-v2/md-combiner-13ed30aa2dc5fda1/lib/python3.13/site-packages/mistletoe/markdown_renderer.py", line 112, in __init__
    block_token.remove_token(block_token.Footnote)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cyberwizzard/.cache/uv/environments-v2/md-combiner-13ed30aa2dc5fda1/lib/python3.13/site-packages/mistletoe/block_token.py", line 62, in remove_token
    _token_types.remove(token_cls)
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
ValueError: list.remove(x): x not in list
```

When I comment out the line stating `block_token.remove_token(block_token.Footnote)` it works. Can it be that I instantiate the class multiple times but the _token_types is a shared list?

Perhaps a check is needed on _token_types to only remove if the type is still present?

gaoflow:
Fixes #262.

### Problem

Instantiating `MarkdownRenderer()` a second time without using it as a context manager crashes:

```python
MarkdownRenderer()
MarkdownRenderer()   # ValueError: list.remove(x): x not in list
```

`MarkdownRenderer.__init__` calls `block_token.remove_token(block_token.Footnote)`, which mutates the module-global `block_token._token_types`. That global is only restored by `reset_tokens()`, which runs in `BaseRenderer.__exit__` — i.e. only when the renderer is used as a `with` context manager. Without `with`, the second instantiation reaches `_token_types.remove(Footnote)` when `Footnote` is already gone and raises. `span_token.remove_token` has the same unguarded pattern.

### Fix

Make `remove_token` idempotent in both `block_token` and `span_token`: only remove the class when it is actually registered. As @pbodnar noted on the issue, a presence check on `_token_types` is what's needed. Context-manager usage is unchanged (verified `reset_tokens` still restores the exact token list on exit).

### Tests

Added a test that instantiates `MarkdownRenderer` twice without a context manager (with cleanup that restores the globals). Without the fix it raises the `ValueError`; with it the full suite passes (342 tests).

Disclosure: I used AI assistance (Claude) to help locate and draft this fix, under my direction and review.

### DIFF
--- a/mistletoe/block_token.py
+++ b/mistletoe/block_token.py
@@ -56,10 +56,15 @@ def remove_token(token_cls):
     Allows external manipulation of the parsing process.
     This function is usually called in BaseRenderer.__exit__.
 
+    Does nothing if the token class is not currently registered, so that it is
+    safe to call repeatedly (e.g. when a renderer that removes a token is
+    instantiated more than once without being used as a context manager).
+
     Arguments:
         token_cls (BlockToken): token to be removed from the parsing process.
     """
-    _token_types.remove(token_cls)
+    if token_cls in _token_types:
+        _token_types.remove(token_cls)
 
 
 def reset_tokens():
--- a/mistletoe/span_token.py
+++ b/mistletoe/span_token.py
@@ -47,10 +47,14 @@ def remove_token(token_cls):
     Allows external manipulation of the parsing process.
     This function is called in BaseRenderer.__exit__.
 
+    Does nothing if the token class is not currently registered, so that it is
+    safe to call repeatedly.
+
     Arguments:
         token_cls (SpanToken): token to be removed from the parsing process.
     """
-    _token_types.remove(token_cls)
+    if token_cls in _token_types:
+        _token_types.remove(token_cls)
 
 
 def reset_tokens():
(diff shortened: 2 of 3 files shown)


pbodnar:
See #281.

PR #283, opened 2026-07-05
Escape mailto autolink hrefs

CodingFeng101:
## What changed

Email autolink `href` values now go through the same URL and HTML escaping path as normal autolinks.

## Why

`render_auto_link()` escaped regular URL targets, but the `mailto:` branch directly interpolated `mailto:{target}` into an HTML attribute. Email local-parts can contain `&`, so input like `<a&b@example.com>` rendered a bare `&` in `href` even though the link text was escaped:

```html
<a href="mailto:a&b@example.com">a&amp;b@example.com</a>
```

After this change, the attribute renders as:

```html
<a href="mailto:a&amp;b@example.com">a&amp;b@example.com</a>
```

## Validation

- Reproduced the pre-fix output locally with `HtmlRenderer().render(Document('<a&b@example.com>'))`.
- `python -m pytest test\test_html_renderer.py::TestHtmlRenderer::test_email_autolink_escapes_href test\test_html_renderer.py::TestHtmlRendererEscaping::test_email_autolink_escapes_href -q`
- `python -m pytest test\test_html_renderer.py -q`
- `python -m pytest test -q`
- `python -m py_compile mistletoe\html_renderer.py test\test_html_renderer.py`
- `git diff --check`

### DIFF
--- a/mistletoe/html_renderer.py
+++ b/mistletoe/html_renderer.py
@@ -94,7 +94,7 @@ class HtmlRenderer(BaseRenderer):
     def render_auto_link(self, token: span_token.AutoLink) -> str:
         template = '<a href="{target}">{inner}</a>'
         if token.mailto:
-            target = 'mailto:{}'.format(token.target)
+            target = self.escape_url('mailto:{}'.format(token.target))
         else:
             target = self.escape_url(token.target)
         inner = self.render_inner(token)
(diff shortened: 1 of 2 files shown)


pbodnar:
@CodingFeng101, thanks for the fix. I can see a slight duplicity in the unit tests, but whatever, I think it is good-enough, merging. :)

PR #289, opened 2026-08-22
Add test_remove_span_from_processing for coverage of span_token

APS1990A:
I was looking at test coverages and saw we were not stepping in a branch that ran `remove_tokens()`. I wrote a small test case for that in `test_span_token.py`.

### DIFF
--- a/test/test_block_token.py
+++ b/test/test_block_token.py
@@ -737,3 +737,8 @@ class TestRemoveToken(unittest.TestCase):
         block_token.remove_token(block_token.BlockCode)
         self.assertNotIn(block_token.BlockCode, block_token._token_types)
 
+        # Check second removal does not raise error (pull request #281 issue #262)
+        try:
+            block_token.remove_token(block_token.BlockCode)
+        except ValueError:
+            self.fail("Idempotent removal failed")
--- a/test/test_block_token.py
+++ b/test/test_block_token.py
@@ -725,6 +725,7 @@ class TestFileWrapper(unittest.TestCase):
         wrapper.reset()
         assert next(wrapper) == "somewhat interesting\n"
 
+
 class TestRemoveToken(unittest.TestCase):
     def setUp(self):
         self.addCleanup(block_token.reset_tokens)
--- a/test/test_block_token.py
+++ b/test/test_block_token.py
@@ -724,3 +724,15 @@ class TestFileWrapper(unittest.TestCase):
         assert next(wrapper) == "somewhat interesting\n"
         wrapper.reset()
         assert next(wrapper) == "somewhat interesting\n"
+
+class TestRemoveToken(unittest.TestCase):
+    def setUp(self):
+        self.addCleanup(block_token.reset_tokens)
+
+    def test_remove_token_from_processing(self):
+        # Check that token exists
+        self.assertIn(block_token.BlockCode, block_token._token_types)
+        
+        block_token.remove_token(block_token.BlockCode)
+        self.assertNotIn(block_token.BlockCode, block_token._token_types)
+
--- a/test/test_span_token.py
+++ b/test/test_span_token.py
@@ -234,12 +234,13 @@ class TestHtmlSpan(unittest.TestCase):
 
 class TestRemoveToken(unittest.TestCase):
     def setUp(self):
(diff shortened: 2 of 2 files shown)


test/test_span_token.py
```diff
@@ -230,3 +230,16 @@ def test_parse_with_illegal_whitespace(self):
...
+    def setUp(self):
+        self.addCleanup(span_token.reset_tokens)
+
+    def test_remove_span_from_processing(self):
+        self.assertEqual(len(span_token._token_types), len(span_token.__all__))
+        span_token.remove_token(span_token.RawText)
+        self.assertEqual(len(span_token._token_types), len(span_token.__all__) - 1)
+        # Check that trying to remove twice still works
+        span_token.remove_token(span_token.RawText)
+        self.assertEqual(len(span_token._token_types), len(span_token.__all__) - 1)
```
pbodnar: A few suggestions:

1. Are we removing a "span", or rather a "token"?
2. Test's dependency on `span_token.__all__` is not actually necessary? (related problem: we try not to repeat ourselves in the tests as well)
3. Before removing the selected token class, let's also check it is actually present in the collection. This will increase the "assumptions section" lines of the test - and we could mark these lines with a comment and a blank line below.

pbodnar (changes requested):
@APS1990A, thanks for this PR, small but still valid. Do you think you could also add explicit, analogical test for the `block_token` (where we've got the removal line covered, but only indirectly)?

Also note, that this generally means writing unit tests for the recent #281 then. :)

APS1990A:
> @APS1990A, thanks for this PR, small but still valid. Do you think you could also add explicit, analogical test for the `block_token` (where we've got the removal line covered, but only indirectly)?

> Also note, that this generally means writing unit tests for the recent #281 then. :)

Yes to both, I will spend some time exploring those. I can open a separate pull request to #281 if that makes sense to you.

test/test_span_token.py
APS1990A: 1. If this question is because the function name is a bit confusing, yes, I think I will change it to match the function `remove_token`. I might have become a bit confused writing the test case with names.
2. I see that you use `_token_types` directly in parts of this test class? I can rewrite the test to use that instead of using `__all__`.
3. Similar to 2, if I use `_token_types` I can add to that list with the token class I want to use.

:edit: I made changes based on the feedback here as well as reviewing some of the other test cases. I will work on the other test cases soon. Let me know if you have feedback.

APS1990A:
> Also note, that this generally means writing unit tests for the recent #281 then. :)

I took a better look at the issue today. I think that my original test already would have been testing that, although indirectly.

 I am not sure if the python `unittest` library supports checking that no exception was thrown, but I could re-write it so it would be more explicit in testing for idempotence.

test/test_span_token.py
pbodnar: @APS1990A, thanks for the amendments, you nailed it. 👍

test/test_block_token.py
```diff
@@ -724,3 +724,21 @@ def test_anchor_reset(self):
...
+        self.assertIn(block_token.BlockCode, block_token._token_types)
+        
+        block_token.remove_token(block_token.BlockCode)
+        self.assertNotIn(block_token.BlockCode, block_token._token_types)
+
+        # Check second removal does not raise error (pull request #281 issue #262)
+        try:
+            block_token.remove_token(block_token.BlockCode)
+        except ValueError:
+            self.fail("Idempotent removal failed")
```
pbodnar, line 744: I think why not, i.e. we can fail explicitly here. I would just simplify the comment:

```suggestion
        # Check second removal does not raise error (see issue #262)
        try:
            block_token.remove_token(block_token.BlockCode)
        except ValueError:
            self.fail("Idempotent removal failed")
```

pbodnar (changes requested):
Thank you, just one last little suggestion.

APS1990A:
@pbodnar  I am not sure why github is still is displaying `Requested changes` but I believe we resolved all of them. If you have anything other requests let me know.

pbodnar:
@APS1990A, thank you, slightly amended and merged. :)

PR #291, opened 2026-08-28
test: remove parameterized dependency

### LINKED ISSUE #258: `parameterized` does not work with modern Pytest
The `parameterized` does not work with Pytest > 4 (see https://github.com/wolever/parameterized/issues/34) and so it would be great if `mistletoe` tests are changed to do not depend on it.  Please also read https://github.com/release-engineering/dist-git/issues/87.

Thank you.

amandeavor:
## Summary

- remove `parameterized` from the test requirements
- convert the seven decorated test methods (20 cases) to standard-library `unittest.subTest` loops
- preserve named case isolation without requiring a third-party parameterization package

`parameterized.expand` still passes under the current Pytest 9.1.1 in my baseline run, so this does not claim a present upstream failure. The change instead resolves the packaging concern in #258 by removing the dependency entirely while keeping the existing unittest-style suite.

## Validation

- Pytest 9.1.1 with `parameterized` uninstalled: 334 passed, 1 skipped, 23 subtests passed
- `python -m unittest discover`: 335 passed, 1 skipped
- CommonMark conformance suite: all tests passing
- critical flake8 checks (`E9,F63,F7,F82,W605`): clean

Fixes #258.


pbodnar:
@amandeavor, thank you, I would wait with this for now though and probably go with another solution - see my analysis at the parent issue.

In short, as you yourself write, `@parameterized.expand` works just fine. Also, I'm not a big fan of the unittest's `subTest()` boilerplate code.

Issue #230, opened 2024-11-01
Add rendering support for Atlassian Document Format


danarth:
It would be really great to be able to parse Markdown and output a JSON document as [Atlassian Document Format][1].

Mistletoe already supports rendering Jira syntax, but ADF is supported in both Jira and Confluence so would be incredibly useful for writing Confluence pages as Markdown.

I guess it could work similarly to the `ASTRenderer` works, since the `BaseRenderer` methods all expect string outputs, whereas this output is a JSON format so could make more sense to keep the responses as dictionaries and convert to a string at the very end.. but keen to hear thoughts.

I'd like to take a crack at this - my time is limited so may take a while :)

[1]: https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/

pbodnar:
PS: I suggest to place the new renderer to the [contrib](https://github.com/miyuchina/mistletoe/tree/v1.4.0/mistletoe/contrib) folder.

APS1990A:
If this is still something we would like to add to the project, I would like try. It is tricky getting the relationship correct between ADF, AST, and the rest of the codebase.

PR #295, opened 2026-09-16
[DRAFT] feat: Add support for ADF renderer

APS1990A:
Issue described here #230.

I will continue to flesh out the renderer and hopefully get some thing releasable and reviewable soon.


APS1990A:
I still want to write tests for this, but I think there is enough to review now. Most cases I have tried with the [ADF Viewer](https://developer.atlassian.com/cloud/jira/platform/apis/document/viewer/) have worked, but it would be helpful if anyone described anymore.

There are a few limits to this: Images do not work (they need some ID generated by Atlassian to work and render properly), and nesting things like block quotes do not work correctly in ADF.

I also would like to rewrite some of this into smaller functions.

APS1990A:
`List` and `ListItem` have no way of determining if a `ListItem` is a checkbox, which means I can not support ADF check boxes. 

The syntax like:
- [x] Some text
- [ ] Other text

I would like to change `List` to add another regex to see if the list is a checkbox list, something like:
`checkbox_pattern = re.compile(r' {0,3}(?:\d{0,9}[.)]|[+\-*]) (\[[x | ]\])?')`

But it might be outside of this pull request and I can open another if that is preferred.

